PR #14, opened 2024-02-29
docs/customcommands: document custom commands section

SoggySaussages:
- docs/customcommands: add custom-commands chapter
- docs/customcommands: document commands page

I plan on adding the database page and proofreading everything in full, as well as manually breaking the lines at 120 characters (VSC user :p) but I'd appreciate feedback to know if I'm heading in the right direction. 

For your convenience I am hosting this branch on https://yagdocsv2.vedamaharaj.ca/custom-commands.html.

**Terms**


content/custom-commands/commands.md
```diff
@@ -0,0 +1,314 @@
...
+
+- There is a syntax error in the response
+- You have reached the maximum CC limit
+- You are attempting to save an empty response
+
+If you save a command with an interval trigger which has never been run, it will run instantly upon saving.
+
+{{% notice style="tip" title="Keeping your code safe" %}}
+
+It is reccomended to code your custom command using a local editor on your device. You will not be able to save your code on the dashboard if there are syntax errors in your code. Use an editor like **Vim**, **VSC**, or **Notepad++** for the best coding experience.
```
savage4618, line 302: consider expanding `VSC` to `VS Code`

savage4618 (commented):
I have a slight complaint about the images looking like they were grabbed from mobile. The more I think about it though, the more I think it might be fine with the amount of people asking about how to navigate the control panel via mobile.

content/custom-commands/commands.md
```diff
@@ -0,0 +1,314 @@
...
+
+Whitelisting roles or channels causes them to be required to run the command.
+
+Blacklisted roles or channels will not be permitted to run the command.
+
+{{% notice style="note" %}}
+
+YAGPDB was raised well and honors a "no" when told "no". In other words, blacklisted roles take precedence over whitelisted roles.
+
+This is a relatively common trip-hazard, so take great care when you set up both whitelisted and blacklisted roles, and subsequently give those roles to your members.
```
l-zeuch, line 67: This feels confusingly worded given how you note blacklisted channels
*and* blacklisted roles — I understand that you copied this from
botlabs-gg/yagpdb-docs-v2#13, but we might want to reword this to also
include the mention about roles, or be generic about a blacklist.

As a tangential note, command overrides as seen in core do not have
blacklisted channels, so this rewording does not exactly apply there.

content/custom-commands/commands.md
```diff
@@ -0,0 +1,314 @@
...
+
+Example:
+
+```go
+  {{ if eq .Reaction.Emoji.APIName "😀" "🦆" }}
+    This is an allowed reaction!
+  {{ else }}
+    This is not an allowed reaction.
+  {{ end }}
+```
```
l-zeuch, line 160: Should disallow :duck:. :stuck_out_tongue:

l-zeuch (changes requested):
You have a somewhat inconsistent use of title case in your headers.

I understand that this may cause inconsistencies with the control panel
due to them being labelled in "normal" case, but I think we should stick
to title case as long as said label isn't a full sentence or something
to that extend.

I.e. "Output errors as command response" stays as-is,
but "Saving your command" should be "Saving Your Command" instead.

.vscode/settings.json
l-zeuch, line 1: Accidental check-out of this file?

I should probably add it to `.gitignore`...

content/custom-commands/_index.md
```diff
@@ -0,0 +1,9 @@
++++
+archetype = "chapter"
+title = "Custom Commands"
+weight = 3
++++
+
+Custom Commands (CCs) allow you to develop your own commands and behaviours for your server. A command may be as simple as an autoresponder or employ a more complex system. A variety of triggers are available for custom commands and the custom template script provides a wide array of templates and functions for coding advanced commands.
```
l-zeuch, line 7: Remember to hard-break at 120 characters. Also, as much as it pains me,
too, we're using American English in this documentation.

content/custom-commands/commands.md
```diff
@@ -0,0 +1,367 @@
...
+command by clicking on it, allowing you to preview the full, syntax-highlighted command response.
+
+### Delete a Command
+
+Deleting a custom command (**4**) will **permanently** delete the command after confirmation. This cannot be undone.
+
+### Run Now
+
+The Run now button (**5**) appears on [Interval trigger](#hourlyminute-interval) commands. It executes the command as
+long as it isn't disabled and a channel is selected, and redirects you to the  edit command page for the command.
```
l-zeuch, line 58: It seems like there are too many consecutive spaces here.
```suggestion
long as it isn't disabled and a channel is selected, and redirects you to the edit command page for the command.
```

content/custom-commands/commands.md
```diff
@@ -0,0 +1,367 @@
...
+{{% notice info %}}
+
+You must specify a channel to run interval commands in even if the command doesn't output a message.
+
+{{% /notice %}}
+
+#### Case Sensitivity
+
+Any commands which allow you to specify trigger text (ex. Command, Regex, Exact match, etc.) have a **Case sensitivity**
+toggle (**5**) which is off by default. A case sensitive trigger `yagPDB` will trigger on "yagPDB" but not "yagpdb" or
```
l-zeuch, line 235: The adjective is normally spelt "case-sensitive"; "Case sensitivity" as
seen on the control panel is also correct, however.

content/custom-commands/commands.md
```diff
@@ -0,0 +1,367 @@
...
+"YAGPDB".
+
+#### Edit Message Trigger
+
+Commands which trigger on messages have a **Trigger on message edits** toggle (**6**) which is off by default. If a
+message is edited and matches the trigger text, it will trigger the command.
+
+The edited message toggle is an _additional_ trigger to the sent message trigger. If you'd like to _only_ trigger on
```
l-zeuch, line 243: Personally, I would just call it a "normal message trigger", "sent
message trigger" feels quite confusing, especially when we take into
account what it is called on the control panel (no such thing).

content/custom-commands/commands.md
```diff
@@ -0,0 +1,367 @@
...
+message edits, you will need to conditional branch with `{{ .IsMessageEdit }}`.
+
+This feature is [premium only](/premium).
```
l-zeuch, line 246: Not entirely sure if it might be worthwhile to move this one up a bit,
perhaps right underneath this very header?

content/custom-commands/commands.md
```diff
@@ -0,0 +1,367 @@
...
+A UTC Timestamp of the last time the command executed the response.
+
+#### Next Scheduled Run
+
+Only shown on Interval type commands. A UTC Timestamp of the next time the command is scheduled to run.
+
+### Output errors as command response
+
+This toggle (**12**) determines whether errors during command execution are sent in the command response after the
+command fails. Does not effect logging of Last Error to the statistics.
```
l-zeuch, line 330:
```suggestion
command fails. Does not affect logging of Last Error to the statistics.
```

content/custom-commands/database.md
```diff
@@ -0,0 +1,80 @@
...
+Value **8** Size of Value **9** Delete Entry **10** Page Navigation **11** Search Query **12** Search Bar
+
+</center>
+
+## Entries
+
+### Entry ID
+
+The ID is assigned by the system and unique within YAGPDB's database. The database page is sorted by descending ID. You
+may seach for a specific ID using the **Search Bar** (**11**) with ID **Query** (**10**).
```
l-zeuch, line 25:
```suggestion
may search for a specific ID using the **Search Bar** (**11**) with ID **Query** (**10**).
```

content/custom-commands/database.md
```diff
@@ -0,0 +1,80 @@
...
+The user-defined key of the entry. A key is a string, max 256 characters. You may search for a specific key using the
+**Search Bar** (**11**) with Key **Query** (**10**).
+
+### Value
+
+The serialized value of the database entry.
+
+### Size of Value
+
+Size of the value in bytes. YAGPDB database entries have a max value size of 100kB.
```
l-zeuch, line 56: You generally have a (non-breaking) space between the numerical value
and the unit symbol, minus a very few exceptions like `°C`. My
professors also always chastize me when I do not do that, so I feel your
pain. :-)
```suggestion
Size of the value in bytes. YAGPDB database entries have a max value size of 100 kB.
```

l-zeuch (changes requested):

.vscode/settings.json
SoggySaussages: I swear I amended this commit like two or three times to get rid of this damn file, and yet >:(

l-zeuch:
I've gone ahead fixed the last two review comments myself.

Thanks for the PR!

PR #26, opened 2024-06-06
notifications: copy current notif docs

l-zeuch:
Copy the currently notifications docs; internal discussion concluded
that this is the current milestone, such that we have a faster migration
to the statically built site.

Some modifications such as notice shortcodes were necessary for it to
work at all. I also took the liberty and removed links to dead /
redundant / not yet existing links.

**Terms**


SoggySaussages:
I've been linking the not yet existing links in my commits thus far, should I stop?

l-zeuch:
Nah it's alright. I just removed those because I wasn't sure where they'll reside in the final doc.

Come to think of it, having dead links makes Hugo emit a warning during build, so it might be actually sensible to just keep these dead links.

PR #28, opened 2024-06-07
reference: migrate reference

SoggySaussages:
Migrate reference from old docs. I've simplified all the tables except for a few raw HTML tables in functions that I'm not
touching with a 10 foot pole. There were some neat tabs on old regex page I don't know how to migrate, so I've simplified
them to headers. All the images referenced are migrated too. I've tried to
catch as many of the hard coded reference links as I could, but a few of the ones in templates and functions tables may
have slipped by. 

**Terms**


l-zeuch:
Thanks!

Re: Emojis (as per our quick chat) — I'll keep them in for now, we can later decide (democratically) whether to remove them or not.

PR #30, opened 2024-06-09
core: add v2.38.0 cmd docs

SoggySaussages:
Generate and add cmd docs from a clean 2.38.0 yag instance.

**Terms**

### DIFF
--- a/content/core/all-commands.md
+++ b/content/core/all-commands.md
@@ -1,0 +1,1442 @@
++++
+title = 'All Commands'
+weight = 4
++++
+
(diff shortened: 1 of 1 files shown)


l-zeuch:
Thanks!

l-zeuch:
Drive-by comment:
Perhaps we can find a way to automate this with GitHub Actions
triggering on a new YAGPDB release, which then re-generates the command
docs and submits a PR if something has changed; we might have to add
this to upstream, though, rather than having it in the docs repository.

PR #31, opened 2024-06-09
cc/interactions: document interactions

SoggySaussages:
Add interaction triggers to commands page
Add interaction templates to templates page
Add interaction functions and alter complexMessage
Add full reference guide on interactions

**Terms**


content/custom-commands/commands.md
```diff
@@ -239,6 +239,38 @@ You must specify a channel to run interval commands in even if the command doesn
...
+
+Use of interactions within YAGPDB is an advanced topic, the documentation should be used only as reference. To learn
+about using interactions, [see here](/reference/custom_interactions).
+
+{{% /notice %}}
+
+The component trigger is used to trigger custom commands via buttons or select menus.
+
+When creating a button or a select menu using template scripting, you must define a `custom_id`. When that button/menu
+is used by a member of your server, it triggers any custom command whose trigger matches that Custom ID.
```
l-zeuch, line 254:
```suggestion
```
I would just leave this sentence out entirely. It is, in spirit, a Write
Everything Twice (WET). You already explained in the cookbook that a
custom ID must be defined for a component to be able to trigger a custom
command.

Perhaps just hint at a custom_id regex being required, if this suggestion
is too DRY for you.

content/custom-commands/commands.md
```diff
@@ -239,6 +239,38 @@ You must specify a channel to run interval commands in even if the command doesn
...
+
+Use of interactions within YAGPDB is an advanced topic, the documentation should be used only as reference. To learn
+about using interactions, [see here](/reference/custom_interactions).
+
+{{% /notice %}}
+
+The modal trigger is used to trigger custom commands via submitting a modal.
+
+When responding to an interaction with a modal using template scripting, you must define a `custom_id`. When that modal
+is submitted by a member of your server, it triggers any custom command whose trigger matches that Custom ID.
```
l-zeuch, line 270: See my comment on the similar section under Component.

content/custom-commands/commands.md
```diff
@@ -239,6 +239,38 @@ You must specify a channel to run interval commands in even if the command doesn
...
+The trigger is matched using [RegEx](/reference/regex).
+
+#### Modal
+
+{{% notice style="tip" %}}
+
+Use of interactions within YAGPDB is an advanced topic, the documentation should be used only as reference. To learn
+about using interactions, [see here](/reference/custom_interactions).
+
+{{% /notice %}}
```
l-zeuch, line 265: This feels extremely repetitive to me — I understand that the current
layout doesn't give us much to work with; what do you think of
summarising these two headers under a collective "Interactions" header?

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,221 @@
...
+{{ sendMessage nil $message }}
+```
+
+![Many buttons](many_buttons.png)
+
+At this stage we have three buttons. Both of the first two buttons will trigger our duck trigger custom command, but the
+third button will not trigger any custom command. Link buttons do not create *interactions*.
+
+We can differentiate between the two buttons with `.StrippedID`, which returns the `.CustomID` but with the trigger and
+everything before it (in our case, `buttons-duck`) stripped off.
```
l-zeuch, line 154:
```suggestion
We can differentiate between the two buttons using `.StrippedID`, which, just like `.StrippedMsg`, returns our Custom ID
without the trigger and everything else before that. In our example, `.StrippedID` will return `-alpha` for the first
button and `-beta` for the second button.

Confirming this behavior will be left as an exercise to the reader (you).
```

What do you think about this rephrase?

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,221 @@
...
+but now since we have two components whose stripped IDs are `-alpha`, we'll need to branch with `.IsMenu` too. If our
+custom command was triggered by the use of this select menu, we could also see that `.MenuType` returns `"string"`,
+since this menu is a string select type menu.
+
+##### Ordering Message Components
+
+A message can have 5 rows of components.
+A row of components can have 5 buttons, **or** 1 menu.
+
+Let's say I wanted to play tic tac toe. If I just add 9 buttons into the same slice in my complex message, they'll just
```
l-zeuch, line 189:
```suggestion
Let's say we want to play tic tac toe. If we just add 9 buttons into the same slice in my complex message, they'll just
```

I'm assuming this a cookbook style you're aiming for, "we" seems more appropriate.

l-zeuch (changes requested):
I understand this PR is still marked as draft, which is why I'd like to
start with the overall flow of the page; we can obsess over little
details later. For now, let's Get Stuff Done.

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,221 @@
...
+Buttons, Modals, Select Menus, Ephemeral Messages, and more!
+
+<!--more-->
+
+{{% notice warning %}}
+
+Use of interactions within YAGPDB is an advanced topic. You will need a thorough understanding of YAGPDB's scripting
+language before learning interactions.
+
+We have a comprehensive learning course located [here](https://learn.yagpdb.xyz) for you to get started.
```
jo3-l, line 15: Wanted to note this down as a general style suggestion: we should avoid using "click here" or "located here" in links; see, e.g., <https://ux.stackexchange.com/questions/12100/why-shouldnt-we-use-words-such-as-here-and-this-in-textlinks> for reasoning. In this case I would prefer to write
```suggestion
We have a [comprehensive learning course](https://learn.yagpdb.xyz) for you to get started.
```

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,221 @@
...
+
+{{% /notice %}}
+
+## The Basics
+
+Interactions within Discord allow server members to use alternative, built-in features to trigger bots to take action
+other than messages or reactions. These features include builtin buttons, dropdown selection menus, or submitting a
+modal (basically a pop-up form). Within custom commands it is possible to both create and customize these new
+interactive features, but respond to them as well, opening up new possibilities for ephemeral message responses, modals,
```
jo3-l, line 24:
```suggestion
Interactions within Discord allow server members to use alternative, built-in features to trigger bots to take action
other than messages or reactions. These features include builtin buttons, dropdown selection menus, or submitting a
modal (basically a pop-up form). Within custom commands it is possible to not only create and customize these new
interactive features, but respond to them as well, opening up new possibilities for ephemeral message responses, modals,
```

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Definitions
+
+**Interaction** - A user engaging with YAGPDB through one of Discord's builtin features: Clicking a button, Making a
+selection with a select menu, or Submitting a modal.
+
+**Response** - YAGPDB is required to send an initial interaction response after receiving an *interaction*. If it does
+not do this, the user triggering the interaction will see a "This application did not respond" error. The bot cannot
+respond to an interaction more than once. YAGPDB may only respond to an interaction in one of the following ways:
+Sending a message, sending a modal, or updating the message which the *interaction* was triggered on.
```
jo3-l, line 35:
```suggestion
**Response** - YAGPDB is required to respond promptly after receiving an interaction by either sending a message or modal, or by updating the message on which the interaction was triggered. If it does not do this, the user triggering the interaction will see a "This application did not respond" error. The bot cannot respond to an interaction more than once.
```
This reformulation puts the definition of "initial interaction response" up front, which I think is preferable.

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,221 @@
...
+
+**Followup** - Since YAGPDB may only *respond* to an *interaction* once, it is subsequently required to send an interaction
+followup if it still needs to respond to the interaction. These followups can be sent up to 15 minutes after the initial
+interaction, and you can send as many as you want. YAGPDB may only send a followup in one of the following ways: Sending
+a message, editing a response or followup, or getting a response or followup message.
+
```
jo3-l, line 41: This is confusingly worded. "YAGPDB may only send a followup in one of the following ways: [...] editing a followup, or getting [...] a followup" (?)

content/custom-commands/commands.md
```diff
@@ -239,6 +239,38 @@ You must specify a channel to run interval commands in even if the command doesn
 
 {{% /notice %}}
 
+#### Component
+
+{{% notice style="tip" %}}
+
+Use of interactions within YAGPDB is an advanced topic, the documentation should be used only as reference. To learn
```
jo3-l, line 246:
```suggestion
Use of interactions within YAGPDB is an advanced topic; the documentation should be used only as reference. To learn
```
nit but I did preface this review with non-blocking ;-)

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,221 @@
...
+**Modal** - A pop-up form YAGPDB can send in response to an interaction. It allows users to privately input text which
+is sent directly to YAGPDB for use in CC scripting. You can create and customize these modals' appearance and
+behavior with a title and fields. YAGPDB can both **receive a submitted modal** (which is an
+*interaction*), and **send a modal** for a member to fill out, (which is an interaction *response*).
+
+**Ephemeral** - An ephemeral message is sent to a server channel but only appears to a single user. YAGPDB cannot send
+these ephemeral messages to users except in response to an *interaction*. Both *response* messages and *followup*
+messages can be ephemeral.
+
+## Interaction Lifetime
```
jo3-l, line 64: This is a nice summary (with a visual as well!) I would suggest it be placed before the definition as a sort of "basic intro". In general I believe guides or tutorials (which this article appears to be, as opposed to reference documentation per se) should begin with a brief overview that still has some usable technical detail before diving into specific definitions and so on.

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,221 @@
...
+Sending a message, sending a modal, or updating the message which the *interaction* was triggered on.
+
+**Followup** - Since YAGPDB may only *respond* to an *interaction* once, it is subsequently required to send an interaction
+followup if it still needs to respond to the interaction. These followups can be sent up to 15 minutes after the initial
+interaction, and you can send as many as you want. YAGPDB may only send a followup in one of the following ways: Sending
+a message, editing a response or followup, or getting a response or followup message.
+
+**Interactive Elements** - Elements users can interact with to send *interactions*, i.e. buttons, menus, and modals.
+
+**Message Components** - *Interactive Elements* which can be attached to YAGPDB's Discord messages, i.e. buttons and menus.
```
jo3-l, line 44: I understand these are the terms that the Discord API uses, but I would rather inline these terms into their use sites. So instead of saying "Creating message components" I would much rather see "Creating buttons and menus"; this seems more amenable to skimming (and, in general, more appropriate for an audience that does not work with the Discord API directly.)

jo3-l (commented):
non-blocking

content/reference/custom_interactions.md
SoggySaussages: Maybe 'editing/getting a previous followup message'?

content/reference/custom_interactions.md
SoggySaussages: I'm trying to get people familiar with the idea of message components for a few reasons:

1. A component of a message makes it clear that these are just parts of any regular message, not just interaction responses. During prototyping a few users struggled to understand how to create buttons or menus in the first place.
2. The CC trigger type is a "message component" type.

l-zeuch: I think this is fine if we write the other reference pages in a similar style, where appropriate.
Otherwise I agree with Joe -- in-depth explanations belong, in my opinion, to learn.yagpdb.

For now I propose we keep this as-is and then once we migrate the learning resources over to Hugo as well we move details over there. What do you think?

content/reference/custom_interactions.md
l-zeuch: Thanks, in my initial review I somehow missed that. I think we have a lot of similar things in the migrated documentation files as well; I will review those at a later date.

content/custom-commands/commands.md
SoggySaussages: Hmm, could we remove the warnings and turn them into hyperlinks "In-depth Interactions Guide"?

l-zeuch: Sure, that sounds like it could work.

l-zeuch (changes requested):
Thanks for the changes, the critical section in the last review reads
much, much smoother now.

Following the feedback for your latest changes.

content/reference/custom_interactions.md
```diff
@@ -138,7 +138,136 @@ This button will now trigger the following custom command:
...
+  "max_values" 3 }}
+
+{{ sendMessage nil (complexMessage "menus" $menu) }}
+```
+
+![The above mentionable menu with default options selected](mentionable_menu_with_defaults.png)
+
+{{% notice info %}}
+
+The `default_values` `id`s must be `strings`, not `int64s`. Note how in the above example, `$adminRoleID :=
```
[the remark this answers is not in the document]
SoggySaussages: I totally agree with this and actually did this initially but thought that might not read well and did it this way instead :ye: Thanks for affirmation that that is the correct way to go

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,725 @@
...
+```
+
+![Result of the Full Solution Code](full-variable-row-solution.png)
+
+#### Emojis in Message Components
+
+Buttons and Select Menu Options both have an `"emoji"` field, but this field does not accept the regular unicode/name:id
+formula like reactions do. Emojis in components follow the [partial emoji
+object](https://discord.com/developers/docs/resources/emoji#emoji-object) structure, however only the ID *or* the Name
+fields are required, depending on if you are using a custom emoji or not.
```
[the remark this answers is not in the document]
SoggySaussages: Yes I will update my code examples to reflect that only an ID is
required when using Custom IDs. I have verified through testing
that this is the case, only an ID is needed for custom emojis and
regardless of if it is animated or not the animated field is not
necessary.

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,725 @@
...
+
+[Interaction Function documentation](/reference/templates/functions#interactions)
+
+### Following Up
+
+Followups allow you to continue responding to an interaction after the initial response has been made. You can followup
+for up to 15 minutes after the user interacts, and you can follow up as many times as you'd like. Followups require the
+interaction token of the interaction they should be following up on.
+
+Possible initial responses:
```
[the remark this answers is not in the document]
SoggySaussages: Sorry that was just me forgetting to edit the title after copy pasting from
the previous section, thank you for catching that.

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,725 @@
...
+
+## Responding to an Interaction
+
+### Initial Response
+
+While technically not required, responding to an interaction with one of Discord's allotted initial responses is crucial
+if you don't want your users to see an error after interacting. An interaction may be responded to one time.
+
+You can only respond to an interaction within the custom command triggered by said interaction. A CC executed with
+`execCC` by the triggered CC will be able to send initial responses to the triggering interaction.
```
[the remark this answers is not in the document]
SoggySaussages: Does this change adequately address?

content/reference/custom_interactions.md
```diff
@@ -0,0 +1,723 @@
...
+
+## Responding to an Interaction
+
+### Initial Response
+
+While technically not required, responding to an interaction with one of Discord's allotted initial responses is crucial
+if you don't want your users to see an error after interacting. An interaction may be responded to only once.
+
+You can only respond to an interaction within the custom command triggered by said interaction, with the exception that
+a CC executed with `execCC` by the triggered CC will be able to send initial responses to the triggering interaction.
```
l-zeuch, line 651:
```suggestion
a CC executed with `execCC` by the triggered CC will be able to send initial responses to the triggering interaction as well.
```

l-zeuch (changes requested):
Everything looks good, just one final thing.

l-zeuch (approved):

jo3-l:
This PR needs to be rebased given the just-merged migration to Doks; happy to help with that if needed.

SoggySaussages:
Blocker: I need to edit a screenshot to remove identifying details of testers

jo3-l (approved):

PR #41, opened 2024-06-23
learn: add learning resources

l-zeuch:
This PR marks the migration of the learning resources currently hosted
with GitBook to Doks, as well as integration into this repository.

I'll be working through files individually and commit them as I go,
rather than one gigantic commit. That way, it is easier to track the
progress and provide interim reviews.

If you do intend on submitting such reviews, review the commits
individually and **do not** review 1277343, as that is just me adding the
current GitBook to this repository.

I might take liberty in merging some files together, to provide a better
flow, especially on (currently) very short pages; some files I might
delete because I determined them to be redundant. If you object to some
of that, do let me know by reviewing the relevant commit.

**Terms**


l-zeuch:
With the latest commit 5c3b47a ("learn: migrate (read: rewrite) inputs 1")
I think the Beginner chapter should be fully migrated to Doks.

I am requesting a review of my changes before I continue working on
reworking the remainder of learn — I want to make sure I'm on the right
track.

To note, I don't see much value in just keeping learn as-is; some of the
written content is outdated, and the structure is not as clear as it
could be. I think it would be beneficial to rework the content to be
more concise and clear (as I am already doing), thereby avoiding the
additional work later on, when we would have merged learn as-is.

savage4618 (approved):

content/learn/README.md
savage4618, line 1: Was this rewritten somewhere else? *I should probably read the rest of everything before posting this question...*

l-zeuch: This is now in `content/learn/welcome/introduction.md`.

content/learn/welcome/introduction.md
```diff
@@ -0,0 +1,41 @@
...
+## General Tips
+
+As you start your journey with YAGPDB, there are a few things you should keep in mind. These tips will help you avoid
+common pitfalls and make your developing experience as a whole more enjoyable.
+
+You should start writing and testing custom commands on a separate testing server on Discord before adding them to your
+main server. This way, you already avoid a lot of problems caused by members potentially interfering with an incomplete
+custom command.
+
+There is no "undo"-button, and kicking YAGPDB won't reset your configuration. Write your changes down first, then you
```
savage4618, line 26: I don't like the hyphen between "undo" and "button". I think it's unnecessary. ¯\_(ツ)_/¯

content/learn/welcome/introduction.md
```diff
@@ -0,0 +1,41 @@
...
+can revert them if something goes wrong.
+
+Never assume YAGPDB is running 24/7/365. It is a free service, and not guaranteed to be up all the time. Do not
+write your custom commands based on that assumption.
+
+Start simple. Don't try to write a complex custom command right away. Start with something simple, then iteratively
+improve upon that.
+
+Be sure to read the [reference documentation](/docs/reference/templates) along the lessons here. We will not cover every
```
savage4618, line 35: I think there's a missing word here:
"Be sure to read the reference docs along WITH the lessons here.

content/learn/welcome/introduction.md
```diff
@@ -0,0 +1,41 @@
...
+detail of every function in these lessons, rather give you the tools necessary to build complex systems.
+
+Over time, your custom commands may get more lengthy and complex. It is a good idea to write them locally in some text
+editor or IDE, and then paste them into the web interface.
```
savage4618, line 39: Maybe link to popular IDEs or at least define `IDE`. If they are a beginner here, maybe they are a beginner programmer, and these "big words" will turn them off to continuing.

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,131 @@
...
+Later on, we may wish to reassign a new value to a variable. We can do this by using the re-assignment operator `=`.
+When and why this becomes necessary will be discussed in a later chapter.
+
+When debugging your code, you might have to figure out the type of a certain variable. You can do this by using the
+`printf` function with the `%T` verb, like so:
+
+```go
+{{ $name := "Alice" }}
+{{ printf "%T" $name }}
+```
```
savage4618, line 133: If you're going to introduce `printf` here, either link to further documentation on it, or allude to discussing it more in depth in a later chapter. As it sits, this page feels incomplete, it left me wanting more.

content/learn/beginner/inputs-1.md
```diff
@@ -0,0 +1,169 @@
...
+member: {{ ($args.Get 1).Nick }}
+```
+
+There is one major things to note about this code: we're starting to repeat a lot of our `$args.Get N` calls! Let's fix
+that first.
+
+```go
+{{ $args := parseArgs 2 "" (carg "int" "coolness level") (carg "member" "target member") }}
+
+{{ if or (gt (toInt $coolness) 100) (lt (toInt $coolness) 0) }}
```
savage4618, line 112: `Responses: template: :3: undefined variable "$coolness"`

savage4618 (changes requested):

content/learn/welcome/introduction.md
```diff
@@ -0,0 +1,41 @@
...
+> The first thing you gotta discard, whenever you're learning about something, is your ego. - Luke Smith
+
+These are the learning resources for YAGPDB's powerful custom command system. This is your playground!
+
+We are going to teach you how to read code as well as how to write it. After all, how can you **write** if you cannot
+**read**? Learn to read, learn to solve problems, and learn to enjoy the process. The community is there to help, as
+long as you put in some effort yourself.
+
+To understand the custom commands interface you will be interacting with on the control panel, please read the
+documentation on [custom commands](/docs/custom-commands/commands).
```
jo3-l, line 15: "These guides assume that you are already familiar with the custom command UI, so, for instance, you know how to add and edit commands through the control panel. Consult the [documentation on custom commands] if unsure."

content/learn/welcome/introduction.md
```diff
@@ -0,0 +1,41 @@
...
+
+## General Tips
+
+As you start your journey with YAGPDB, there are a few things you should keep in mind. These tips will help you avoid
+common pitfalls and make your developing experience as a whole more enjoyable.
+
+You should start writing and testing custom commands on a separate testing server on Discord before adding them to your
+main server. This way, you already avoid a lot of problems caused by members potentially interfering with an incomplete
+custom command.
```
jo3-l, line 24: "This avoids any potential issues caused by members interfering with incomplete custom commands."

content/learn/welcome/introduction.md
```diff
@@ -0,0 +1,41 @@
...
+can revert them if something goes wrong.
+
+Never assume YAGPDB is running 24/7/365. It is a free service, and not guaranteed to be up all the time. Do not
+write your custom commands based on that assumption.
+
+Start simple. Don't try to write a complex custom command right away. Start with something simple, then iteratively
+improve upon that.
+
+Be sure to read the [reference documentation](/docs/reference/templates) along the lessons here. We will not cover every
+detail of every function in these lessons, rather give you the tools necessary to build complex systems.
```
jo3-l, line 36: "; rather, we aim to give you ..."

The current wording reads a little strangely to me.

content/learn/welcome/introduction.md
```diff
@@ -0,0 +1,41 @@
...
+custom command.
+
+There is no "undo"-button, and kicking YAGPDB won't reset your configuration. Write your changes down first, then you
+can revert them if something goes wrong.
+
+Never assume YAGPDB is running 24/7/365. It is a free service, and not guaranteed to be up all the time. Do not
+write your custom commands based on that assumption.
+
+Start simple. Don't try to write a complex custom command right away. Start with something simple, then iteratively
+improve upon that.
```
jo3-l, line 33: "Don't try to write a complex custom command right away: instead, start with an initial prototype and iteratively improve on it."

The repetition of "start simple" is unfortunate, and can be avoided.

content/learn/welcome/introduction.md
```diff
@@ -0,0 +1,41 @@
...
+
+You should start writing and testing custom commands on a separate testing server on Discord before adding them to your
+main server. This way, you already avoid a lot of problems caused by members potentially interfering with an incomplete
+custom command.
+
+There is no "undo"-button, and kicking YAGPDB won't reset your configuration. Write your changes down first, then you
+can revert them if something goes wrong.
+
+Never assume YAGPDB is running 24/7/365. It is a free service, and not guaranteed to be up all the time. Do not
+write your custom commands based on that assumption.
```
jo3-l, line 30: "Never assume that YAGPDB is running 24/7/365: we do our best, but occasional downtime—for routine maintenance, upstream issues, and otherwise—is inevitable."

Premium exists, so the free service bit is not totally true. "Assumption" is also repeated.

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,85 @@
...
+title = "Outputs 1"
+weight = 210
++++
+
+In this chapter, we will go over two ways to output text from a custom command: using the *response*, and later on using
+*template actions*.
+
+## Response
+
+If you paste text into the response field of a custom command, the bot will respond with that text when the command is
```
jo3-l, line 11: "with the same text"

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,85 @@
...
+triggered.
+
+For instance, paste the following text into a new custom command:
```
jo3-l, line 14: "create a new custom command with the following response:"

avoids the duplication of "paste".

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,85 @@
...
+![Hello World example ran in Discord.](hello_world.png)
+
+</center>
+
+## Template Actions
+
+### Actions As Output
+
+Earlier, we just made the bot respond with some static text. However, sometimes (or rather, most of the time) we want to
+change that around depending on various factors. This is where *template actions* come in.
```
jo3-l, line 33: "However, often we want our custom command to behave differently depending on input. This is where ..."

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,85 @@
...
+
+Template actions are a way to dynamically change the output depending on various things, such as the user who triggered
+the command, the arguments passed to the command, or even the current time.
```
jo3-l, line 36: "depending on environmental data"

I don't particularly like the word "environmental", but something better than "various things" is needed.

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+As you can see, it differs from the static response in that it uses double curly braces (`{{` and `}}`) to denote an
+action. Without those, the bot would just respond with the text as-is.
+
+If we want to bring this a step further, we can combine the plain response with some template actions to make it a bit
+more nicer-looking:
+
+```go
+Hey there {{.User.Username}}!
```
jo3-l, line 52: nit: space

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,85 @@
...
+change that around depending on various factors. This is where *template actions* come in.
+
+Template actions are a way to dynamically change the output depending on various things, such as the user who triggered
+the command, the arguments passed to the command, or even the current time.
+
+The bot evaluates the template action and replaces it with the result of the evaluation. For instance, the following
+code will make the bot respond with the server name when the command is triggered:
+
+```go
+{{ .Server.Name }}
```
jo3-l, line 42: would rather use .Guild here

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+As you can see, it differs from the static response in that it uses double curly braces (`{{` and `}}`) to denote an
+action. Without those, the bot would just respond with the text as-is.
```
jo3-l, line 46: "The double curly braces (`{{` and `}}`) are essential: without these, the bot would simply respond with the text verbatim."

It may be worth adding a separate example illustrating this common pitfall, e.g., a command with the response `Hello .User.Username`.

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,85 @@
...
+
+```go
+Hey there {{.User.Username}}!
+Welcome to {{ .Server.Name }}!
+```
+
+Play around with this a little bit and see what you can come up with. Take a look at the
+[data reference documentation](/docs/reference/templates/syntax-and-data) to see what other data you can access.
+
+### Actions For Computation
```
jo3-l, line 59: I wonder if there is a framing of this introduction that allows a smoother transition to actions that, well, simply perform actions. For instance, `addRoleID` adds a role but is not a computation or a task per se.

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,85 @@
...
+
+Template actions can also be used to perform computations (tasks) and return the result. This allows you to create
+extremely powerful custom commands that do much more than just output some random text. For instance, the following
```
jo3-l, line 62: "do much more than simply output text"

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,85 @@
...
+code will make the bot respond with the sum of two numbers when the command is triggered:
+
+```go
+{{ add 5 3 }}
+```
+
+Obviously this is a very simple example. We provide a list of all available functions in the
```
jo3-l, line 69: "contrived example"

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,131 @@
++++
+title = "Data Types 1"
+weight = 220
++++
+
+In this chapter, we will go over the basic data types that are available in the bot's template actions. These data types
+are used to store various types of data, such as numbers, strings, and more. We will also go over how to use variables
```
jo3-l, line 7: "We will overview how to store data in variables and the basic data types available, which include numbers, strings, and more."

I think presenting data types before data is debatable pedagogy. Data first seems more reasonable to me.

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,131 @@
...
+numbers together, but you cannot add two strings together (at least not in the way you might expect).
+
+## Strings
+
+A string is a sequence of characters. You can think of it as a word or a sentence. In the bot's template actions, you
+can create a string by enclosing some text in double quotes (`"`). For instance, `"Hello, world!"` is a string. We call
+those *quoted strings*.
+
+Now, here we might run into a problem quite quickly: what if we want to include a double quote in our string? We can't
+just write `"Peter said "Hello, world!""`, as the bot would interpret this in a nonsensical way. To solve this, we
```
jo3-l, line 25: "as the bot would not know where to end the string"

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+For a full list of escape sequences, you can refer to the [Go documentation](https://golang.org/ref/spec#Rune_literals).
+Please note that not all escape sequences are supported by Discord.
+
+### Raw String Literals
+
+It should become relatively clear that a lot of new lines and other special characters can make a quoted string quite
+hard to read. To make this easier, you can use backticks (`` ` ``) to create a *raw string literal*. This enables you to
+*literally* press enter to insert a new line in your string, without having to use the `\n` escape sequence, like so:
```
jo3-l, line 45: "A raw string literal does not attempt to interpret its contents in any way, and will simply contain the text between the opening \` and the closing \` unmodified."

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,131 @@
...
+
+```txt
+`This is my
+cool multiline string!
+
+Look at all the new lines!`
+```
+
+To insert a backtick in a raw string literal, however, you must use its ASCII code, which is `\x60`. Escape sequences as
```
jo3-l, line 54: This statement is not correct. The resulting string will contain the sequence of characters `\x60` verbatim (though this is sufficient to match a backtick character in a regular expression, which is the probable cause of confusion.)

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,131 @@
...
+decimal point). In the bot's template actions, you can create an integer by simply writing a number, such as `5`. For
+floating-point numbers, you can add a decimal point, like so: `5.0`.
+
+### Integers
+
+In the bot's template actions, integers are represented as 64-bit signed integers. This means that you can store numbers
+from `-9223372036854775808` to `9223372036854775807`. If you try to store a number outside this range, the bot will
+return an error.
+
+There are several ways to define an integer, specifically:
```
jo3-l, line 69: The bot accepts several notations for integers:

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,131 @@
...
+
+1. As a base-10 number, such as `42`. This will mean what you think, the number forty-two.
+2. As a base-16 number, such as `0x2A`. This is the [hexadecimal representation][hex] of the number forty-two.
+3. As a base-8 number, such as `052`. This is the [octal representation][oct] of the number forty-two.
```
jo3-l, line 73: I would suggest using the `0o` prefix instead, which Go supports and is less likely to cause confusion. (Many newer languages only support `0o` octal literals for this reason.)

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,131 @@
...
+There are a lot of ways to define a floating-point number, but the most common way is to use the decimal point, such as
+`3.14`. For a full list of ways to define a floating-point number, you can refer to the
+[Go documentation](https://golang.org/ref/spec#Floating-point_literals).
+
+### Imaginary Numbers
+
+Go (the language the bot is written in) also supports imaginary numbers (and by extension complex numbers).
+
+Defining them is quite similar to the formats discussed above, but with the addition of the imaginary unit `i`.
+Please refer to the [Go documentation](https://golang.org/ref/spec#Imaginary_literals) for more information.
```
jo3-l, line 95: I would prefer to remove this section entirely. It is wholly out of place in an introductory document and better relegated to some deep, dusty reference section.

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Booleans
+
+A boolean is a data type that can only have one of two values: `true` or `false`. Booleans are used to represent the
+truth value of an expression. For instance, `5 > 3` would evaluate to `true`, while `5 < 3` would evaluate to `false`.
+
+You can think of it as a light switch: it can either be on (`true`) or off (`false`). Booleans are used in conditional
+statements, such as `if` statements, to determine which branch of the code should be executed.
+
+## Variables
```
jo3-l, line 105: As mentioned previously, I would rather introduce variables before data types.

jo3-l (commented):
various nits, will review the remaining commits later

content/learn/beginner/outputs-1.md
l-zeuch: `addRoleID` is a task though—"add the role with this ID".

I do agree however, "Computation" is a bit weird, I had a mental note to change this to something better as well.

content/learn/beginner/datatypes-1.md
l-zeuch: Certainly agreed, I'm not too sure what I was thinking when I did that.

content/learn/beginner/outputs-1.md
l-zeuch: Maybe `Actions For Functionality`? But that also reads like :sparkles: hot garbage :sparkles:

content/learn/beginner/control-flow-1.md
```diff
@@ -0,0 +1,213 @@
...
+- `le` (less than or equal to `<=`)
+- `gt` (greater than `>`)
+- `ge` (greater than or equal to `>=`)
+
+These functions can only compare basic data types as introduced in [Data Types 1](/learn/beginner/datatypes-1). Strings
+are compared using the [underlying Unicode values](https://en.wikipedia.org/wiki/List_of_Unicode_characters), stepping
+through the strings byte by byte. Please refer to the [functions documentation](/docs/reference/templates/functions)
+for the syntax of these functions.
+
+### If Statements
```
jo3-l, line 40: I would rather put this section first, noting that `eq` is a comparison function that checks whether its arguments are equal and that the full list of comparison functions will be enumerated later.

content/learn/beginner/control-flow-1.md
```diff
@@ -0,0 +1,213 @@
...
+{{ else }}
+  Five is not equal to three or five!
+{{ end }}
+```
+
+#### Guard Clauses
+
+As your code grows, you may find yourself nesting `if` statements inside each other. This can lead to code that is hard
+to read and understand. One way to avoid this is to use *guard clauses*. A guard clause is an `if` statement that checks
+for a condition and returns early if the condition is false.
```
jo3-l, line 80: We have not introduced `return` yet (I think?), so although the concept is simple it would still be nice to briefly elaborate: "and exits early using the `{{ return }}` action if the condition is false."

content/learn/beginner/control-flow-1.md
```diff
@@ -0,0 +1,213 @@
...
+
+- `eq` (equals `==`)
+- `ne` (not equals `!=`)
+- `lt` (less than `<`)
+- `le` (less than or equal to `<=`)
+- `gt` (greater than `>`)
+- `ge` (greater than or equal to `>=`)
+
+These functions can only compare basic data types as introduced in [Data Types 1](/learn/beginner/datatypes-1). Strings
+are compared using the [underlying Unicode values](https://en.wikipedia.org/wiki/List_of_Unicode_characters), stepping
```
jo3-l, line 36: byte-by-byte contradicts "[comparison] using underlying Unicode values". The comparison is done via (an equivalent process to) memcmp so byte-by-byte is accurate.

content/learn/beginner/control-flow-1.md
```diff
@@ -0,0 +1,213 @@
...
+Until now, we have just written some linear code that executes from top to bottom. However, in real-world applications,
+we often need to make decisions based on certain conditions. This is where *control flow* comes in.
+
+You already have an intuitive understanding of control flow. For instance, when you cross the street, you look left and
+right to see if any cars are coming. If there are no cars, you cross the street. If there are cars, you wait until they
+have passed, then check again. This is a simple example of a decision-making process.
+
+## Comparison Actions
+
+In programming, we can make decisions by comparing values to each other. We can compare numbers, strings, and other data
```
jo3-l, line 15: might be worth mentioning for programmers with prior experience that comparison is done via functions, not binary operators

content/learn/beginner/control-flow-1.md
```diff
@@ -0,0 +1,213 @@
...
+In Custom Commands, you use `:=` to define a variable, and `=` to reassign a variable. The bot will not throw an error
+if you try to re-define a variable using `:=`, but it will not affect the outer scoped variable.
+
+{{< /callout >}}
+
+## Boolean Logic
+
+To form more complex conditional checks, we need to understand *boolean logic*. [Boolean logic] is a branch of algebra
+that deals with true/false values. In programming, we use boolean logic to make decisions based on the truth value of
+expressions. There are three primary boolean operators: `AND`, `OR`, and `NOT`.
```
jo3-l, line 132: Mentioning boolean algebra is misplaced in an introductory document. All the operators have natural explanations: e.g., OR reports whether at least one expression is true: I will be late to school if `(i wake up late) or (the bus is late)`.

content/learn/beginner/control-flow-1.md
```diff
@@ -0,0 +1,213 @@
...
+
+The logical `NOT` operator is the simplest of these three. It takes one argument and gives back the exact opposite of
+whatever you give it. We provide this operator in the form of the `not` function in Custom Commands.
+
+Expressed in a *truth table*, the `NOT` operator looks like this:
+
+| A     | NOT A |
+|-------|-------|
+| false | true  |
+| true  | false |
```
jo3-l, line 146: Drop the truth tables as well, IMO.

content/learn/beginner/inputs-1.md
```diff
@@ -0,0 +1,169 @@
...
+to a custom command. For simple commands, this is the recommended way. In a later chapter, we will explore a more
+hands-on approach to parsing arguments.
+
+### Defining Arguments
+
+The first step is to define the arguments that the command will take. This is done using the aforementioned `parseArgs`
+function. The syntax is as follows:
+
+```go
+{{ $args := parseArgs <"num required args"> <"error message"> ...carg }}
```
jo3-l, line 22: The `<"...">` scheme is quite confusing. How about just `parseArgs required_args err_msg ...cargs`?

content/learn/beginner/inputs-1.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+The first argument `num required args` is the number of required arguments. You can use this to tell `parseArgs` how
+many arguments of the given definitions at the end are required. If the number of defined arguments is greater than the
+number of required arguments, the remaining arguments are considered optional. You cannot define optional positional
+arguments using `parseArgs`.
+
+After that, we can provide a custom error message that will be displayed if the arguments are not valid. Passing an
+empty string `""` will use a generated one based on the defined arguments. This is useful for providing more context to
```
jo3-l, line 31: will generate one based on the argument definitions

content/learn/beginner/inputs-1.md
```diff
@@ -0,0 +1,169 @@
...
+function. The syntax is as follows:
+
+```go
+{{ $args := parseArgs <"num required args"> <"error message"> ...carg }}
+```
+
+The first argument `num required args` is the number of required arguments. You can use this to tell `parseArgs` how
+many arguments of the given definitions at the end are required. If the number of defined arguments is greater than the
+number of required arguments, the remaining arguments are considered optional. You cannot define optional positional
+arguments using `parseArgs`.
```
jo3-l, line 28: I was a little confused reading this paragraph as a whole. Why not leave it at the first sentence (which seems perfectly clear to me -- perhaps a short example to drive home the point at most?) The rest I could not quite grok: "how many arguments of the given definitions at the end are required" seems redundant; the "optional" part follows from `!required`, and the "cannot define optional positional arguments" I cannot make sense of.

content/learn/beginner/inputs-1.md
```diff
@@ -0,0 +1,169 @@
...
+Combining all of this, let's create a custom command that takes two arguments: a coolness level and a user to apply it
+to.
+
+```go
+{{ $args := parseArgs 2 "" (carg "int" "coolness level") (carg "member" "target member") }}
+```
+
+Note that we are using the `member` type here --- in general, it is recommended to use the `member` type for
+user-related arguments, as it validates that the target user is also a member of the server. There are very limited
+use-cases for `user` and `userid`, which will be left as an exercise to the reader.
```
jo3-l: I do not really agree with the "very limited use-cases" part. If one just needs the user object, use a `user` carg. (It will validate that the user is a member of the server as well.) If one needs the member object, on the other hand, use the `member` carg. Finally, if one only needs a general user ID (without regard for whether it is valid), `userid` has the same effect as `int` but is self-documenting.

content/learn/beginner/inputs-1.md
```diff
@@ -0,0 +1,169 @@
...
+### Accessing Arguments
+
+Currently, our code doesn't do anything with the arguments. To access the arguments, we use the `.Get` method on the
+`$args` variable. The syntax is as follows:
+
+```go
+{{ $args.Get <index> }}
+```
+
+The `index` is the index of the argument, starting from 0. The arguments are stored in the order they are defined in the
```
jo3-l, line 78: the position of the argument, starting from 0

content/learn/beginner/inputs-1.md
```diff
@@ -0,0 +1,169 @@
...
+### Validating Arguments
+
+#### Specifying Valid Ranges
+
+Now, we want to limit the coolness level to a number between 0 and 100. We can do this by adding a simple check:
+
+```go
+{{ $args := parseArgs 2 "" (carg "int" "coolness level") (carg "member" "target member") }}
+
+{{ if or (gt (toInt $args.Get 0) 100) (lt (toInt $args.Get 0) 0) }}
```
jo3-l, line 97: this doesn't work correctly; `toInt $args.Get 0` means `toInt ($args.Get) 0`. Also, the guard clause obfuscates rather than helps here, IMO.

content/learn/beginner/inputs-1.md
```diff
@@ -0,0 +1,169 @@
...
+{{ end }}
+
+coolness: {{ $args.Get 0 }}
+member: {{ ($args.Get 1).Nick }}
+```
+
+Now, we can make use of another great feature of `parseArgs`: Certain types support additional arguments that can be
+used to validate the input. For example, the `int` type supports two additional arguments that can be used to specify a
+range of valid values. This is especially useful if there are a lot of values to check, as it reduces the amount of code
+we have to write, thus freeing up this valuable space for other things.
```
jo3-l, line 127: the "especially useful" part is self-evident; I would suggest just saying "can be used to specify a range of valid values, such that the bot will do the validation for us"

content/learn/beginner/control-flow-1.md
l-zeuch: You're quite right, we haven't introduced `return` yet. I think it's
sufficient enough to explain that we use `return` to exit early.
Later on, when we introduce `define` and friends, we can then expand the
definition of return and show how it's used to return a value from a
template.

content/learn/beginner/control-flow-1.md
l-zeuch: To be fair I copied that more or less verbatim from the old files; I was
confused by that as well. I'll take your word for it and *maybe* check
again for myself, just to be sure.

content/learn/beginner/control-flow-1.md
l-zeuch: It was there at the start of the old learn resources 😎. I do agree
though, it's a bit of a jump to go from simple comparisons to full on
boolean algebra. Do we want to place that in a reference document?

content/learn/beginner/inputs-1.md
l-zeuch: Initially I was going for a `<required> [optional]` format, but
evidently I kind of lost track of doing that, as evident by the `...carg`
notation. Your proposed format seems reasonable.

content/learn/beginner/inputs-1.md
l-zeuch: Reading this section back I do agree it is confusingly written and
unnecessarily verbose. I'm not quite sure how to properly introduce the
concept of optional arguments when `num defined args > num required`,
but I'm positive the example further down should illustrate that to a
satisfactory extend.

content/learn/beginner/inputs-1.md
l-zeuch: Will rewrite this section to be more friendly towards `user` and
`userid` type arguments :-)

content/learn/beginner/inputs-1.md
l-zeuch: The guard clause is indeed meant to obfuscate this code a little bit; I
want to show that it is just way easier to use `parseArg`'s validation
feature, which the following paragraph should hopefully illustrate.

content/learn/beginner/control-flow-1.md
```diff
@@ -0,0 +1,213 @@
++++
+title = "Control Flow 1"
+weight = 230
++++
+
+Until now, we have just written some linear code that executes from top to bottom. However, in real-world applications,
+we often need to make decisions based on certain conditions. This is where *control flow* comes in.
```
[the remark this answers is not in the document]
l-zeuch: I have rewritten this to read as follows:
> we often need to execute different code depending on certain states of our program.

Let me know if it now reads better.

l-zeuch:
I have removed the section on Boolean Algebra in Control Flow 1 for now
and added it to a locally persistent file. Will add it back once I get
to a later section, or the reference section in the main documentation.

content/learn/welcome/introduction.md
```diff
@@ -41,3 +41,24 @@ editor or IDE, and then paste them into the web interface. Some popular choices
...
 [Notepad++]: https://notepad-plus-plus.org/
+
+### Planning
+
+Writing code is (or at least, should be), 90% planning. Investing an extra 10 minutes in carefully planning out a piece
+of code can save hours of debugging a snarled mess later on. Many novice programmers mistakenly jump right into writing
+code without a plan, only to end up pouring hours into what should be a relatively short task.
+
+Planning first is not only the best approach for novices, but also for skilled programmers. However, if you see a highly
+experienced programmer in action, you may not see her planning when working on a relatively easy problem. Not seeing
```
jo3-l, line 52:
```suggestion
experienced programmer in action, you may not see them planning when working on a relatively easy problem. Not seeing
```

jo3-l (commented):
til i am a novice programmer 😔 😔

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
++++
+title = "Outputs 2"
+weight = 310
++++
+
+Until now, we have just used the [default response behavior](/learn/beginner/outputs-1) to make our custom commands
+respond with some text. This makes sense for quick mockups or relatively simple commands. However, this may be
+inconvenient or Not What You Want in some cases. In this chapter, we will explore how to send messages to different
+channels, edit already sent messages, and send messages with embeds, vastly expanding your toolbox for creating complex
```
jo3-l, line 9: edit existing messages?

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+between them is how they handle special mentions like `@everyone` and `@here`, and whether to return the ID of the
+message after sending it. We will cover these functions in detail in the following sections, and when to best use them.
+
+Let's get started with the simplest of them all, `sendMessage`. Its syntax is the following:
+
+```go
+{{ sendMessage channel_id message_to_be_sent }}
+```
+
+The `channel_id` is the ID of the channel in which the message is to be sent. If you want to send the message in the
```
jo3-l, line 24: ID of the channel to send the message to

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+
+```go
+{{ sendMessageNoEscape channel_id message_to_be_sent }}
+```
+
+### Returning the Message ID
+
+If you want to store the ID of the message you just sent, for example to later edit it, you use the `sendMessageRetID`
+function. Assign the result of this function to a variable, otherwise it will be rendered to the output and become
+unusable for further usage in your custom command.
```
jo3-l, line 44: for example to later edit it, use the `sendMessageRetID` function and assign the result into a variable.

The second sentence is probably superfluous.

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+
+```go
+{{ $messageID := sendMessageRetID channel_id message_to_be_sent }}
+```
+
+Naturally, we provide a combination of the `sendMessageRetID` function with the `sendMessageNoEscape` function if you
+want to send a message with special mentions and store the ID of the message.
```
jo3-l, line 51: we provide a variant of the `sendMessageRetID` function that does not escape mentions:

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+
+It is quite the mouthful to write, but just roll with it.
+
+## Building Embeds
+
+Embeds are a powerful way to display information in a structured and visually appealing way. They can contain a title,
+description, fields, images, and more. We provide a function to build embeds in a way that closely resembles the
+structure as defined by the Discord API.
+
+We will illustrate this with a sparse example. For a full breakdown of all available fields, please refer to our
```
jo3-l, line 65: with a simple example

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+    )
+    "author" (sdict
+      "name" .User.Username
+      "icon_url" (.User.AvatarURL "256")
+    )
+}}
+{{ sendMessage nil $embed }}
+```
+
+<div style="float: right; margin-left: 7px; margin-bottom: 10px;">
```
jo3-l, line 85: Is the raw HTML strictly necessary? I'd prefer to leave it out if possible.

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+The above code will generate an embed as shown on the left. Let us dissect it a bit. First, we simply define a variable
+`$embed` and assign it the result of the `cembed` function. This function takes a series of key-value pairs, where the
+key is the field you want to set, and the value is the value you want to set for that field. The `cembed` function will
+return a structured object that can be sent as a message, as demonstrated in the last line of the code.
+
+The `"title"` and `"description"` fields are self-explanatory---we can use Discord Markdown in the latter. The `"color"`
+field is in our case a hexadecimal color value that will be used as the color of the embed, but it can also take a
+[decimal value](https://www.binaryhexconverter.com/hex-to-decimal-converter).
+
+The `"fields"` field is a list (we call it slice) of dictionaries, where each dictionary represents a field in the embed.
```
jo3-l, line 100: a list (more precisely a _slice_)

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+## Editing Messages
+
+Sending a message is nice and all, but for the sake of keeping things clean, you might want to edit a message instead of
+creating a new one each time something changes. We provide the `editMessage` function for this purpose.
+
+```go
+{{ editMessage channel_id message_id new_message_content }}
+```
+
+The `channel_id` is the ID of the channel in which the message is located. The `message_id` is the ID of the message you
```
jo3-l, line 128: ID of the channel containing the message

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+want to edit. The `new_message_content` is the new content you want to set for the message, completely replacing it.
+You should note that YAGPDB can only edit messages from itself, just like you cannot edit someone else's messages.
```
jo3-l, line 130: is the new content of the message, which will completely overwrite the existing content. See [editing-embeds] if you only want to change certain fields of an existing embed while leaving others intact.

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+This code sends a message saying "Hello, World!" and then, after 5 seconds, edits the message to say "Goodbye, World!",
+all in the same channel where the custom command was triggered.
+
+### Editing Embeds
+
+Editing embeds is a little more involved than editing regular messages. You need to provide the entire embed object
+again, old parts included. This is due to how the Discord API functions. The basic idea is that we obtain the old embed
+object, modify it, and then send the modified object back to the `editMessage` function.
```
jo3-l, line 147: Editing embeds is a little more involved than editing regular messages. Since the data provided to `editMessage` completely overwrites the existing content, it is necessary to retrieve the existing embed object, modify the desired fields, and provide the whole embed to `editMessage`.

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+
+A contrived example all within the same custom command looks like the following:
```
jo3-l, line 149: An elaborate example

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+      "icon_url" (.User.AvatarURL "256")
+    )
+}}
+{{ $messageID := sendMessageRetID nil $embed }}
+{{ sleep 5 }}
+
+{{ $new_embed := structToSdict $embed }}
+{{ $new_embed.Set "title" "This is a new title" }}
+{{ $new_embed.Set "description" "This is a new description." }}
+{{ $new_embed.Set "fields" (cslice) }}
```
jo3-l, line 171: Not sure if I like this. One might rightfully question why `$new_embed.Del "fields"` is insufficient (and the answer would be that `$new_embed.Del "Fields"` is necessary due to `structToSdict` not converting case), but that complicates the explanation considerably. Might be worth leaving this line out and addressing it later in the section discussing deep conversion? If you choose to keep it in, explain why `$new_embed.Del "fields"` doesn't work.

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+    )
+    "author" (sdict
+      "name" .User.Username
+      "icon_url" (.User.AvatarURL "256")
+    )
+}}
+{{ $messageID := sendMessageRetID nil $embed }}
+{{ sleep 5 }}
+
+{{ $new_embed := structToSdict $embed }}
```
jo3-l, line 168: newEmbed

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+{{ $new_embed.Set "fields" (cslice) }}
+{{ editMessage nil $messageID (cembed $new_embed) }}
+```
+
+In the second part, after `{{ sleep 5 }}`, we first convert the original embed object to a dictionary using the
+`structToSdict` function. We then modify the fields we want to change using the `Set` method. In this case, we change
+the title and description of the embed, and remove all fields. Finally, we send the modified embed object back to the
+`editMessage` function.
+
+As mentioned previously, this example is contrived and meant to demonstrate the process. In practice, we're more likely
```
jo3-l, line 180: this example is contrived: in practice, we're more likely to ...

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+to use `.Message.Embeds` to get the original embed object, convert and modify it as shown above, then send it back to
+the `editMessage` function.
+
+{{< callout context="note" title="Note" icon="outline/info-circle" >}}
+
+`structToSdict` does not perform deep conversion and has some other general inconsistencies. For a full conversion of an
```
jo3-l, line 186: `structToSdict` does not perform deep conversion and does not convert the case of field names.

(structToSdict isn't inconsistent; it's perfectly well-defined--but it isn't sufficient for the use-case.)

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,249 @@
...
+}}
+
+{{ sendMessage nil $message }}
+```
+
+Similarly, we provide a `complexMessageEdit` function to edit messages with both content and embeds. The syntax is the
+same as `complexMessage`, minus a few keys that cannot be edited. Please refer to the documentation below.
+
+The `complexMessage` builder takes a lot more keys that give you fine-grained control over the message you want to send.
+For a full breakdown of all available keys, please refer to the [message functions documentation][message-docs].
```
jo3-l, line 246: Refer to the [message functions documentation] for a complete description.

jo3-l (commented):
review 1/n

content/learn/intermediate/data-types-2.md
```diff
@@ -1,178 +1,70 @@
...
+
+When you first started this course, you learned about [primitive data types](/learn/beginner/datatypes-1). In the
+previous chapter we name-dropped two more types: "slice", and "sdict" ("map"), both without much elaboration.
+In this chapter, we will explore these data types in more detail.
+
+## Slices
+
+Think of a slice as a list of items, that you can add to or remove from. You might have already heard of an array, which
+is a fixed-size list of items. A slice is like an array, but it can grow or shrink in size. This makes slices more
+powerful than a simple array.
```
jo3-l, line 14: Custom commands do not work with arrays much so while the explanation of a slice as a dynamic array is accurate, I don't know if it's particularly helpful. What about just:

"A slice is an ordered list of items that you can add to or remove from."

l-zeuch:
With [a4f0807](https://github.com/botlabs-gg/yagpdb-docs-v2/pull/41/commits/a4f08075cbdfbe21df4e8f4ed23ec0390bc61121), I will now mark this PR as ready for a review, though we should keep nitpicking purely out of pragmatism of getting this shipped for later.

content/learn/intermediate/data-types-2.md
```diff
@@ -1,178 +1,70 @@
...
+
+In custom commands, we can **c**reate a **slice** with the `cslice` function, like so:
```
jo3-l, line 16: we can create a slice by providing the elements in order to the `cslice` function

content/learn/intermediate/data-types-2.md
```diff
@@ -1,178 +1,70 @@
...
+
+```go
+{{ $mySlice := cslice 1 2 3 "hello" 4 5 "good bye" }}
+```
+
+As you can see, the data types of the elements in a slice need not be equal, they can be anything. You already
```
jo3-l, line 22: As you can see, the data types of the elements in a slice need not be the same: in the above example, we freely mixed integers and strings. Besides primitives, slices can also contain more complex data types: you already subconsciously ...

content/learn/intermediate/outputs-2.md
l-zeuch: The image is rather small—I'd like to keep it to the right of the text. I'm not aware of a markdown / shortcode way.

content/learn/intermediate/data-types-2.md
```diff
@@ -1,178 +1,70 @@
...
+subconsciously used a slice with dictionaries as elements, when you created the `"fields"` slice for an embed in the
+previous chapter.
+
+For available operations on slices, please refer to [our template documentation][docs-slices].
+
+[docs-slices]: /docs/reference/templates/syntax-and-data/#templatesslice
+
+## Maps
+
+A map is an unordered collection of key-value pairs for fast lookups. In custom commands, we have two kinds of maps:
```
jo3-l, line 32: A map is an table that associates values with keys. For example, a program that implements a reputation system might use a map from user IDs to reputation points internally. In custom commands, we have two kinds of maps:

content/learn/intermediate/data-types-2.md
```diff
@@ -1,178 +1,70 @@
...
+For available operations on maps, please refer to [our template documentation][docs-maps].
+
+[docs-maps]: /docs/reference/templates/syntax-and-data/#templatessdict
+
+{{< callout context="tip" title="Here be Dragons" icon="outline/rocket" >}}
+
+Consider the following code:
+
+```go
+{{ $some_map := sdict "key3" "value3" "key1" "value1" "key4" "value4" "key2" "value2" }}
```
jo3-l, line 54: someMap

content/learn/intermediate/outputs-2.md
l-zeuch: Not me mixing camelCase and snake_case

content/learn/intermediate/data-types-2.md
```diff
@@ -1,178 +1,70 @@
...
+Yielding this output:
+
+```txt
+map[key1:value1 key2:value2 key3:value3 key4:value4]
+```
+
+Whilst its output may suggest that `$some_map` ends up ordered by key, this is not actually the case. This behavior is
+implemented for convenient debugging, but is in no way representative of the actual layout in memory.
+Maps are, by definition, unordered. If you expect your map to be ordered in some way, or attempt to sort it, you are
+almost always using the wrong data structure and should consider using a slice instead.
```
jo3-l, line 68: Though the output is ordered by key, this feature is only implemented for debugging convenience; maps are otherwise unordered data structures. If you rely on the order of the entries in a map, consider using a slice of key-value pairs instead.

jo3-l (commented):
review 2/n

content/learn/intermediate/outputs-2.md
l-zeuch: Ah, that was me misinterpreting the `embedtosdict` snippet from the support server, there it says:

```
tructToEmbed does not change the keys so there are some inconsistencies (causing some parts of the embed to be left out without the following code).
```

Certainly well-defined, as you said, just initially unintuitive.

content/learn/intermediate/outputs-2.md
jo3-l: Fair enough, resolving.

content/learn/beginner/control-flow-1.md
```diff
@@ -126,6 +126,23 @@ if you try to re-define a variable using `:=`, but it will not affect the outer
...
 {{< /callout >}}
 
+<div style="float: right; margin-left: 7px;">
+
+![Responses: template: :XX: unexpected EOF](unexpected-eof.png)
+
+</div>
+
+Try and recall the last example from the introduction of if-statements and notice how, despite us having multiple blocks
+of branching code, there's still only one `{{ end }}` action.
```
jo3-l, line 136: I'm not too sure how this sentence is relevant; could you clarify?

content/learn/beginner/control-flow-1.md
```diff
@@ -126,6 +126,23 @@ if you try to re-define a variable using `:=`, but it will not affect the outer
...
+
+Each control structure's start (`if`, `with`, etc.) must eventually have such a closing statement. Not doing so will
+result in an error similar to `template: :XX: unexpected EOF`, where `:XX:` is the line number where the error
+occurred---an example shown above.
```
jo3-l, line 140: Each control structure (`if`, `with`, etc.) must be closed by an `{{end}}` action. If you forget to do so, the custom command engine will not know where your control structure terminates and hence issues the `template: :XX: unexpected EOF` error.

content/learn/beginner/control-flow-1.md
```diff
@@ -126,6 +126,23 @@ if you try to re-define a variable using `:=`, but it will not affect the outer
...
+
+If you're familiar with C-style programming languages, this concept will most likely strike you as the curly braces
+these languages use to denote code blocks.
```
jo3-l, line 143: If you are familiar with C-family programming languages, this error is analogous to missing the closing `}` of a block of code.

content/learn/beginner/control-flow-1.md
l-zeuch:
```
if {
    // code
} else {
    // other code
}
```
Note the additional `{}` for the else block; if we're going to say "yeah this is like the closing curly brace", this "but actually only once" seems relevant to me. Not too sure how to phrase this better, or if it's just better to outright remove it.
But then, I do have seen code like this (please do not ask me when / where):

```
{{ if pipeline }}
    code
{{ end }}
{{ else }}
    more code
{{ end }}
```

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+Play around with this a little bit and see what you can come up with. Take a look at the
+[data reference documentation](/docs/reference/templates/syntax-and-data) to see what other data you can access.
+
+### Actions for Functions
+
+Custom command functions allow you to do various things, such as performing calculations, adding or removing a role from
+a user, or even sending a message to a channel. The syntax is a little different to what you might be used to; the
+function name comes before its arguments, like so:
```
galen8183, line 68:
```suggestion
Custom command functions allow you to perform calculations, add or remove roles to/from
a user, send messages to a channel, and lots more! The syntax is a little different to what you might be used to; All
arguments to a function follow the function name itself, like so:
```
I don't like "do various things."

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,248 @@
...
+
+```go
+{{ sendMessage channel_id message_to_be_sent }}
+```
+
+The `channel_id` is the ID of the channel to send the message to. If you want to send the message in the same channel as
+the one in which the custom command was triggered, simply set `channel_id` to `nil`.
+
+Intuitively, `message_to_be_sent` denotes the output that is to be sent as a message. For now, we will just say it is a
+string with the content you want to send. We will cover sending embeds in a later section on this page.
```
galen8183, line 28:
```suggestion
Intuitively, `message_to_be_sent` denotes the output that is to be sent as a message. For now, we will just use a
string with the content you want to send. We will cover sending embeds in a later section on this page.
```

content/learn/beginner/control-flow-1.md
```diff
@@ -0,0 +1,180 @@
...
+{{ else }}
+  Five is not equal to three or five!
+{{ end }}
+```
+
+#### Guard Clauses
+
+As your code grows, you may find yourself nesting `if` statements inside each other. This can lead to code that is hard
+to read and understand. One way to avoid this is to use _guard clauses_. A guard clause is an `if` statement that checks
+for a condition and returns early via the `{{ return }}` action if the condition is false.
```
galen8183, line 55:
```suggestion
for a condition and returns early via the `{{ return }}` action. Logically this means we need to invert the condition we would have checked with an `if`/`else` statement.
```
We are still checking if a condition is true.

content/learn/beginner/inputs-1.md
```diff
@@ -0,0 +1,165 @@
...
++++
+
+When we were starting to write Custom Commands, we stuck to a somewhat static custom command---one that doesn't take
+any arguments and has a fixed output. This is a good starting point, but it's not very useful in practice. In this
+chapter, we'll learn how to create custom commands that take inputs.
+
+## Parsing Arguments
+
+Because parsing arguments is no easy task, we provide a convenience function, `parseArgs`, to parse the arguments passed
+to a custom command. For simple commands, this is the recommended way. In a later chapter, we will explore a more
```
galen8183, line 13:
```suggestion
to a custom command. For simple commands, this is the recommended way to handle user input. In a later chapter, we will explore a more
```

content/learn/intermediate/control-flow-2.md
```diff
@@ -0,0 +1,268 @@
...
+`.` holds the first element of the slice: `(sdict "Name" "chips" "Calories" 540)`. Hence
+
+```txt
+{{ .Name }} contain {{ .Calories }} calories.
+```
+
+evaluates to
+
+```txt
+chips contains 540 calories.
```
galen8183, line 45:
```suggestion
chips contain 540 calories.
```

content/learn/intermediate/control-flow-2.md
```diff
@@ -0,0 +1,268 @@
...
+### Break and Continue
+
+In custom commands, we provide two actions to control the flow of loops: `{{ break }}` and `{{ continue }}`. `break`
+exits the loop prematurely, whereas `continue` skips the remainder of the current iteration and jumps to the next one.
+These can prove very useful to optimize your code for size and readability, with similar benefits to guard clauses with
+`{{ return }}` introduced in earlier chapters.
+
+## With Blocks
+
+Just like the `if` action, `with` runs a block of code if the given expression is truthy. The only difference is that
```
galen8183, line 240: I don't think truthiness is explained anywhere, this may not be the best place but it should be clarified somewhere.

content/learn/intermediate/data-types-2.md
```diff
@@ -0,0 +1,81 @@
...
+
+yielding
+
+```txt
+map[apple:1.5 banana:2.6 pineapple:3.5]
+```
+
+Though the output is ordered by key, this feature is only implemented for ease for debugging: maps are otherwise
+unordered data structures. If your program relies on the entries of a map being ordered, consider use a slice of
+key-value pairs instead.
```
galen8183, line 79:
```suggestion
unordered data structures. If your program relies on the entries of a map being ordered, consider use a slice of
maps instead.
```

galen8183 (commented):
Mostly minor wording changes and typos

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,126 @@
...
+{{ printf "%T" $name }}
+```
+
+{{< /callout >}}
+
+## Data Types
+
+If you're completely new to programming, you might not know what a data type is. You can think of a data type as a way
+to distinguish between different kinds of "things". As a real-life analogy, you can separate food into several
+categories, such as fruits, vegetables, and meat. Each category is in that sense its own "data type".
```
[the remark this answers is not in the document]
jo3-l: American English puts punctuation inside the quotation mark; I personally dislike this and follow the British convention of putting (non-source) punctuation outside the quotation. That said, in this case I'd drop the quotation marks entirely.

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,248 @@
...
+## Sending Messages
+
+We provide several `sendMessage*` functions that all do the same thing: send a message to a channel. The difference
+between them is how they handle special mentions like `@everyone` and `@here`, and whether to return the ID of the
+message after sending it. We will cover these functions in detail in the following sections, and when to best use them.
+
+Let's get started with the simplest of them all, `sendMessage`. Its syntax is the following:
+
+```go
+{{ sendMessage channel_id message_to_be_sent }}
```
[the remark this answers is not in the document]
jo3-l: The case difference is indeed to differentiate between function names and arguments. I'd prefer to standardize on snake_case so we should probably leave the cases here as is.

content/learn/intermediate/outputs-2.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+The `channel_id` is the ID of the channel to send the message to. If you want to send the message in the same channel as
```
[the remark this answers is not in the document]
jo3-l: Saying that channel_id could be a name of a channel is something I'd rather avoid; we could rename `channel_id` to `channel` or, alternately, keep the text as is and let users figure out that names are also permitted from the reference documentation. The latter is my preference; I don't want to encourage using names in guide-level text which (aims to, at least) illustrate best practices. WDYT?

content/learn/intermediate/control-flow-2.md
```diff
@@ -0,0 +1,138 @@
...
+{{ end }}
+```
+
+This does not apply to the optional `else` block following a `range` action.
+
+{{< /callout >}}
+
+### While
+
+`while` iterates as long as the specified condition is true, or more generally evaluates to a non-empty value. The dot
```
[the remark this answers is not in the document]
jo3-l: text/template doc refer to the term 'empty value', though I standardized on 'truthy value' in the rewrite of this page.

content/learn/intermediate/database.md
```diff
@@ -0,0 +1,399 @@
...
+A common use case for this function is a cooldown: As long as the entry exists, the command is still on cooldown.
+
+```go
+{{if $db := (dbGet 2000 "cooldown")}}
+    Command is on cooldown :(
+    Cooldown will be over at {{$db.ExpiresAt.Format "Mon 02 Jan 15:04:05"}}
+{{else}}
+    Command is not on cooldown :)
+    {{dbSetExpire 2000 "cooldown" "random value" 60}}
+{{end}}
```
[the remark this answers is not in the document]
jo3-l: Good concern re: tab width, but I think we shouldn't block this PR on formatting concerns. Should definitely make a note somewhere though.

content/learn/intermediate/data-types-2.md
jo3-l: Slice of key-value pairs is intentional, though it could be clarified. (Tangential, but now that I read this sentence again, `consider use` is not quite right. Whoops.)

content/learn/intermediate/outputs-2.md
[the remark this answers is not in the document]
galen8183: Fair point, I agree that this should show best practices.

content/learn/beginner/outputs-1.md
```diff
@@ -0,0 +1,90 @@
...
+```go
+{{ mult 5 (add 3 2) }}
+```
+
+Just like in math, the expression inside the parentheses is evaluated first, and then that result is used in the outer
+expression.
+
+Obviously these are both quite contrived examples. We provide a list of all available functions in the
+[functions reference documentation](/docs/reference/templates/functions). Try to experiment around with some of these
+functions to get a feel for how they're called and what they do.
```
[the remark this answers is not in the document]
l-zeuch: I meant "called" as in "making a function *call*"; though probably more / most natural would be "how they're used" in that case.

content/learn/beginner/datatypes-1.md
```diff
@@ -0,0 +1,126 @@
++++
+title = "Data Types 1"
+weight = 220
++++
+
+In this chapter, we will overview how to store data in variables and the basic data types available, which include
+numbers, strings, and more.
```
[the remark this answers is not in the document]
l-zeuch: I think directly linking to the headers is appropriate, will do that.

content/learn/beginner/datatypes-1.md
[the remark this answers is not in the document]
l-zeuch: As much as it pains me, too, we should stick to American English conventions simply for consistency's sake—we're already using the US spelling (e.g. color instead of colour), therefore it doesn't make much sense to suddenly introduce UK conventions on top.


(Already said that on Discord, just regurgitating here for the record)

content/learn/intermediate/control-flow-2.md
jo3-l: Agreed, but finding a home for this information is surprisingly difficult. The most natural location—Control Flow 1 introducing if conditionals—is not totally appropriate, unfortunately, since the complex data types have not been introduced at that stage. I suppose we could only show the boolean case in Control Flow 1 and allude to if conditionals operating on a more general definition of empty values. The empty values for each data type could then be clarified as they are introduced in Data Types 1 and Data Types 2 (so, for instance, after presenting the slice type we mention that an empty slice is falsy.)

A true conundrum indeed.

content/learn/intermediate/outputs-2.md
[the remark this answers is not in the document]
galen8183: Sounds good, the database page extensively uses `<PascalCase>` currently so it should be changed instead.

content/learn/beginner/datatypes-1.md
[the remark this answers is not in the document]
jo3-l: Lord save me from this nightmare

Do not go gentle into that good night,
Rage, rage against the dying of the light.

content/learn/intermediate/data-types-2.md
galen8183: I read "a slice of key-value pairs" as something like `cslice "key" "value"`, maybe this is just me from trying to understand people in support channels though :)

jo3-l: Hmm. I see how "slice of key-value pairs" could be confusing in the context of custom commands where "key-value pairs" is overloaded, but it's fairly common terminology otherwise. I don't particularly like "slice of maps" though... maybe "slice of entries" would be better? Alternately, we could leave it at "use a slice instead".

content/learn/intermediate/control-flow-2.md
l-zeuch: I think interspersing it with `if`-actions and the respective data-type is a sensible approach, because there really isn't any good place to introduce this in a centralised paragraph.
~~I also have the hope that future readers will actually read rather than skim but that's a pipe dream of mine~~

galen8183: That sounds like a good approach to me too, the initial explanation could also be expanded upon to include numbers and strings along with booleans.

content/learn/beginner/control-flow-1.md
```diff
@@ -0,0 +1,181 @@
...
+![Responses: template: :XX: unexpected EOF](unexpected-eof.png)
+
+</div>
+
+Try and recall the last example from the introduction of if-statements and notice how, despite us having multiple blocks
+of branching code, there's still only one `{{ end }}` action.
+
+Each control structure's start (`if`, `with`, etc.) must eventually have such a closing statement. Not doing so will
+result in an error similar to `template: :XX: unexpected EOF`, where `:XX:` is the line number where the error
+occurred---an example shown above.
```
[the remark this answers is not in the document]
l-zeuch: If we're moving the image below this block of text in the raw markdown, it will flow into the next section.

image

Having it in the centre and below the text doesn't feel nice to me, as it is relatively small—in the rendered page it is above and to the right of this sentence.

content/learn/intermediate/database.md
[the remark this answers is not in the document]
l-zeuch: We'll go with four spaces wide indentation going forward; I've made a note to review other code blocks as well with regards to this tab width across all pages.

content/learn/intermediate/database.md
```diff
@@ -0,0 +1,399 @@
...
+{{/* or as query */}}
+{{dbCount (sdict "userID" <userID> "pattern" <pattern>)}}
+```
+
+dbCount counts the entries either for the given database userID or the pattern for database keys. Alternatively, you can
+make it count entries as one query by passing in an `sdict` with the keys `userID` for the ID and `pattern` for database
+keys that are to be counted.
+
+Please note that this function returns the amount of entries, so to avoid random spam, you'd have to store it into a
+variable. Apart from that, this function is very simple to use and might come in handy for a few things.
```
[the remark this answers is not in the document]
l-zeuch: Not necessary at all IMO—we can just clearly state that it returns the number of entries matching the given criteria and leave it at that.

content/learn/intermediate/database.md
```diff
@@ -0,0 +1,399 @@
...
+`dbBottomEntries` in ascending order. You cannot retrieve more than 100 elements in one call - honestly though, 100
+entries in one go is definitely a lot, probably more you would have to actually deal with.
+
+```go
+{{dbTopEntries <pattern> <amount> <nSkip>}}
+{{dbBottomEntries <pattern> <amount> <nSkip>}}
+```
+
+What's new here are three things: `pattern`, `amount`, and `nSkip.` Let's walk through them one by one. For `pattern`,
+we use basic PostgreSQL patterns, you can read on them further down. The `amount` specifies how many entries we want to
```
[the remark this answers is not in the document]
l-zeuch: Definitely should link to the `Patterns` section, not sure whether to actually include the explanation under dbCount, though.
Personally I prefer it as a separate heading, simply to have it out of the way and an easier time linking to it.

content/learn/intermediate/database.md
```diff
@@ -0,0 +1,399 @@
...
+
+As an example, to find the rank of the entry with the key `test` for the current user in all of this user's entries, you
+may want to use the following code:
+
+```go
+{{$rank := dbRank (sdict "userID" .User.ID) .User.ID "test"}}
+The specified entry's rank is {{$rank}}.
+```
+
+Note that this function **returns** the rank, so to avoid random spam, don't forget to assign it to a variable.
```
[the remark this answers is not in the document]
l-zeuch: I think this entire sentence is superfluous, as we're already stating that this function returns something. ~~What was I drinking when I wrote this~~

content/learn/intermediate/database.md
```diff
@@ -0,0 +1,399 @@
...
+
+You might have noticed that, whenever you're storing a user ID, channel ID, etc. into your database, it will come back
+as a weird value, such as `5.241379415938826e+17`. This is because they're saved as floats, hence it being messed up.
+To prevent this, simply convert them to a string before storing and converting back to `int` upon retrieving, like so:
+
+```go
+{{dbSet 2000 "someKey" (str .User.ID)}}
+{{$userID_received := toInt (dbGet 2000 "someKey").Value}}
+{{dbDel 2000 "someKey"}}
+```
```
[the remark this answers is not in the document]
l-zeuch: I think you meant to try and remove `str` (`toString`), rather than `toInt`?

Also, I'm not sure where GitHub messed up with parsing for highlighting purposes, but if I remove that single backtick, the page won't render correctly anymore; highlighting is proper in my editor.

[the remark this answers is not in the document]
galen8183: Yes, you're correct on both counts.
Github was ending the suggestion code block there and it looked quite messy (though I could have omitted that entirely now that I reread)

content/learn/intermediate/database.md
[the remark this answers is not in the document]
galen8183: Sorry, I meant to ask if that section should be introduced and linked under dbCount. The main nit is that `pattern` is not really new.
This is changed in Joe's rewrite anyway to introduce and link to the Patterns section :)

content/learn/beginner/control-flow-1.md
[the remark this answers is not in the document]
galen8183: How about beside the paragraph that references it, rather than the first paragraph? It feels slightly out of place there to me.
2024-07-22T122832

content/learn/intermediate/control-flow-2.md
jo3-l: Done in https://github.com/botlabs-gg/yagpdb-docs-v2/pull/41/commits/916648171a27b96e830eddf631ef4fff5c9a3825.

jo3-l:
OK, that's all remaining review comments addressed. Let's :shipit:!

Thanks @l-zeuch for the absolutely herculean effort here, and thanks also to @savage4618 and @galen8183 for the very thorough review!

PR #53, opened 2024-10-31
fix custom-interactions.md

NakshatraSirohi:
fix ./content/docs/reference/custom-interaction.md -- line 187

before  :  ["opt-1" "duck-option" "opt-3"]
after  :  ["opt-1" "duck-option" "third-option"]


lemm-e:
What is the "fix" here?

PR #55, opened 2024-11-17
reference: improve accuracy of closeThread desc

SoggySaussages:
`closeThread` does not archive a thread if the lock flag is provided. It will do one or the other. Also adds a hint for
users who'd like to do both that they should lock before closing.

**Terms**

Signed-off-by: SoggySaussages <vmdmaharaj@gmail.com>

### DIFF
--- a/content/docs/reference/templates/functions.md
+++ b/content/docs/reference/templates/functions.md
@@ -33,9 +33,9 @@ Adds a member to an existing thread. Does nothing if either argument is invalid.
 {{ closeThread <thread> <lock> }}
 ```
 
-Closes the given thread.
+Closes or locks the given thread. You cannot lock a closed thread, but you can close a locked thread.
 
-- `lock`: whether to also lock the thread. Default `false`.
+- `lock`: whether to lock the thread instead. Default `false`.
 
 #### createForumPost
 


content/docs/reference/templates/functions.md
```diff
@@ -33,9 +33,9 @@ Adds a member to an existing thread. Does nothing if either argument is invalid.
 {{ closeThread <thread> <lock> }}
 ```
 
-Closes the given thread.
+Closes or locks the given thread. You cannot lock a closed thread, but you can close a locked thread.
```
l-zeuch, line 36: This feels awkwardly worded just for the sake of having a brain-teaser. Are we able to formulate this in a clearer, more instructive way, e.g.

```suggestion
Closes or locks the given thread. If you wish to do both, call this function first with the `lock` flag, then again without it to close the thread afterwards.
```

l-zeuch (changes requested):

content/docs/reference/templates/functions.md
SoggySaussages: It wasn’t really for the sake of a brain teaser, more along the lines of ‘this is what this function can and can’t do’ rather than ‘here is how you can accomplish a common task using this function in your code.’ If you’d prefer the latter, sounds good. Perhaps the answer is both? Or just slightly less confusing wording for the ‘can and can’t do’ approach? 

If I make the function do both I need to add a new func for just locking. Seems like a lot of breaking changes when we could just change the docs instead.

l-zeuch: See if you can come up with a clearer "can and cannot" approach, reading my suggestion back it isn't much better than the initial change.
If you cannot, we can ship this as-is.

l-zeuch (approved):

PR #56, opened 2024-11-22
reference/functions: correct carg duration min max

SoggySaussages:
Correct the format the `min` and `max` values for a `duration` type `carg` accept.

**Terms**

### DIFF
--- a/content/docs/reference/templates/functions.md
+++ b/content/docs/reference/templates/functions.md
@@ -2289,7 +2289,7 @@ An argument's `"type"` must be one of the following:
   duration)
 
 Additionally, the `int`, `float`, and `duration` type support validation ranges in the interval `(min, max)`, where for
-`duration` it is in seconds.
+`duration` it is in nanoseconds, or simply pass [time.Duration format](/docs/reference/templates/syntax-and-data#time) values.
 
 ##### Example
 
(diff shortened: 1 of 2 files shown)


content/docs/reference/templates/functions.md
```diff
@@ -2289,7 +2289,7 @@ An argument's `"type"` must be one of the following:
   duration)
 
 Additionally, the `int`, `float`, and `duration` type support validation ranges in the interval `(min, max)`, where for
-`duration` it is in seconds.
+`duration` it is in nanoseconds, or simply pass [time.Duration format](/docs/reference/templates/syntax-and-data#time) values.
```
jo3-l, line 2292: "where for `duration`, the min and max are [time.Duration] values".

I don't know if we want to encourage using nanoseconds anywhere, and even if we do want to include nanoseconds in the documentation, we should put it after mentioning time.Duration since that's preferable.

jo3-l (commented):
Thanks. The corresponding section in learn needs to be fixed too, if that's not too much trouble.

SoggySaussages:
> Thanks. The corresponding section in learn needs to be fixed too, if that's not too much trouble.

donezo

jo3-l (approved):
Thanks!

PR #59, opened 2024-12-01
cc: document crontab trigger

SoggySaussages:
Documents the new crontab trigger, linking robfig's docs, Wikipedia, and outlining some examples of common schedules.

**Terms**

Signed-off-by: SoggySaussages <vmdmaharaj@gmail.com>


content/docs/custom-commands/commands.md
```diff
@@ -242,6 +252,61 @@ The modal trigger is used to trigger custom commands via submitting a modal.
 
 The trigger is matched using [RegEx](/docs/reference/regex).
 
+##### Crontab
+
+This trigger will run the command at scheduled intervals using [Cron Job Scheduling](https://en.wikipedia.org/wiki/Cron). This allows for advanced specificity in scheduling execution, such as scheduling execution at 23:45 every Saturday.
```
jo3-l, line 257: I think we should avoid using the term "scheduled intervals" in this section to avoid confusion with interval triggers. Drawing an explicit contrast to interval triggers would also be nice. I propose

"This trigger runs the command periodically at times specified using [cron syntax]. In contrast to interval triggers, which run a command with a fixed delay but unknown time, cron triggers can be made to execute at fixed times of day. For instance, a cron trigger with the expression `45 23 * * 7` runs every Saturday at 23:45 UTC."

I'm not particularly fond of the phrase "execute at fixed times of day", given that cron triggers really have more flexible applications, but I have nothing better at the moment.

content/docs/custom-commands/commands.md
```diff
@@ -242,6 +252,61 @@ The modal trigger is used to trigger custom commands via submitting a modal.
...
+
+When using an interval trigger, the custom command does not receive any user or member context. Thus, `{{ .User.ID }}`
```
jo3-l, line 259: when using a cron trigger

content/docs/custom-commands/commands.md
```diff
@@ -242,6 +252,61 @@ The modal trigger is used to trigger custom commands via submitting a modal.
...
+
+<center>
+
+**1** Cron Expression **2** Channel **3** Excluding hours/weekdays
+
+</center>
+
+Cron Expression (**1**) defines the expression used to schedule the cron job. It uses the standard expression format (`minute hour day month dayofweek`). It does not support predefined schedules such as `@hourly`. The cron scheduler uses UTC always.
+
+Each field additionally supports special characters: `* / , -`. To read more about use of special characters, visit [Robfig's Cron package documentation - Special Characters](https://pkg.go.dev/github.com/robfig/cron/v3#hdr-Special_Characters).
```
jo3-l, line 272: I wonder if we should copy the relevant snippets of documentation there to save an extra click for interested users, with appropriate credit. WDYT @l-zeuch?

content/docs/custom-commands/commands.md
```diff
@@ -242,6 +252,61 @@ The modal trigger is used to trigger custom commands via submitting a modal.
...
+
+0 0 1 1 *
+Run once a year, midnight, Jan. 1st.
+```
+
+Channel (**2**) specifies a channel to run the command in. The response, if any, will be sent to this channel.
+
+Excluding hours and/or weekdays (**3**) prevents the command from triggering during those hours or weekdays. **This uses UTC time**, not your local timezone.
+
+You must specify a channel to run interval commands in even if the command doesn't output a message.
```
jo3-l, line 302: run cron commands in

content/docs/reference/templates/functions.md
```diff
@@ -2289,7 +2289,7 @@ An argument's `"type"` must be one of the following:
   duration)
 
 Additionally, the `int`, `float`, and `duration` type support validation ranges in the interval `(min, max)`, where for
-`duration` it is in seconds.
+`duration` it is in nanoseconds, or simply pass [time.Duration format](/docs/reference/templates/syntax-and-data#time) values.
```
jo3-l, line 2292: are you based on an old commit? this change should already be on master

jo3-l (changes requested):
As a general comment, I think we should link to https://crontab.guru/ or a similar tool somewhere in this documentation.

content/docs/custom-commands/commands.md
SoggySaussages: I wonder if we could instead change the term to "time trigger" to maintain the umbrella statement encompassing both interval and cron without adding more confusion between cron and interval. Thoughts?

content/docs/reference/templates/functions.md
SoggySaussages: Yes I can rebase.

content/docs/custom-commands/commands.md
l-zeuch: For the record, these "special" characters are part of any decent cron implementation—the `cron` packages does nothing new here and ALLBUT the slash are also in the relevant standard.
This text is, in a similar way, also already on Wikipedia, just quite far down.

Copy from either the cron package or Wikipedia and link to the relevant section.

content/docs/custom-commands/commands.md
```diff
@@ -242,6 +249,95 @@ The modal trigger is used to trigger custom commands via submitting a modal.
...
+<center>
+
+**1** Cron Expression **2** Channel **3** Excluding hours/weekdays
+
+</center>
+
+Cron Expression (**1**) defines the expression used to schedule the cron job. It uses the standard expression format
+(`minute hour day month dayofweek`). It does not support predefined schedules such as `@hourly`. The cron scheduler uses
+UTC always.
+
```
jo3-l, line 273: Can we add a tip callout here referencing https://crontab.guru/?

content/docs/custom-commands/commands.md
```diff
@@ -242,6 +249,95 @@ The modal trigger is used to trigger custom commands via submitting a modal.
...
+To read more about the supported format of cron expressions, visit [Robfig's Cron package documentation - Expression
+Format](https://pkg.go.dev/github.com/robfig/cron/v3#hdr-CRON_Expression_Format).
+
+**Special Characters**
+
+Each field additionally supports special characters: `* / , -`.
```
jo3-l, line 279: The word "additionally" is strange here, because we have not discussed what parameters each field accepts prior to this sentence. I'd rather just include
```text
A cron expression represents a set of times, using 5 space-separated fields.

Field name   | Mandatory? | Allowed values  | Allowed special characters
----------   | ---------- | --------------  | --------------------------
Minutes      | Yes        | 0-59            | * / , -
Hours        | Yes        | 0-23            | * / , -
Day of month | Yes        | 1-31            | * / , - ?
Month        | Yes        | 1-12 or JAN-DEC | * / , -
Day of week  | Yes        | 0-6 or SUN-SAT  | * / , - ?

Month and Day-of-week field values are case insensitive. "SUN", "Sun", and "sun" are equally accepted. 
```
in full before this as well.

jo3-l (changes requested):
Sorry to nitpick, but there are some suggestions remaining from my last review. We should be good to merge after.

jo3-l (approved):
Sorry for the confusion. I just meant the reference to crontab.guru and the (not fully realized) suggestion to copy the documentation from robfig/cron. This LGTM now, but I will wait for @l-zeuch for any final thoughts and merge.

content/docs/custom-commands/commands.md
```diff
@@ -242,6 +249,110 @@ The modal trigger is used to trigger custom commands via submitting a modal.
...
+Month        | Yes        | 1-12 or JAN-DEC | * / , -
+Day of week  | Yes        | 0-6 or SUN-SAT  | * / , - ?
+
+To read more about the supported format of cron expressions, visit [Robfig's Cron package documentation - Expression
+Format](https://pkg.go.dev/github.com/robfig/cron/v3#hdr-CRON_Expression_Format).
+
+{{< callout context="tip" title="Tip: Debugging Cron Expressions" icon="outline/rocket" >}}
+
+To help build and debug cron expressions, we reccomend using [Cronitor's Crontab Guru](https://crontab.guru/) or a
+similar site. Note that predefined schedules such as `@hourly`, and the use of `7` in the DOW field may parse correctly
```
l-zeuch, line 289: You should introduce this abbreviation in the table above; perhaps also for DOM.

content/docs/custom-commands/commands.md
```diff
@@ -242,6 +249,110 @@ The modal trigger is used to trigger custom commands via submitting a modal.
...
+Your cron expression must schedule jobs with greater than a 10 minute interval between executions.
+
+{{< /callout >}}
+
+Channel (**2**) specifies a channel to run the command in. The response, if any, will be sent to this channel.
+
+Excluding hours and/or weekdays (**3**) prevents the command from triggering during those hours or weekdays. **This uses
+UTC time**, not your local timezone.
+
+You must specify a channel to run time-based commands in even if the command doesn't output a message.
```
l-zeuch, line 354: This should be moved to the rest of the image caption at the start of this section.

l-zeuch (changes requested):
Just two stylistic remarks, the rest looks good.

content/docs/custom-commands/commands.md
l-zeuch: @SoggySaussages I cannot see any change addressing this comment--did you perhaps forget to push your changes or publish your response?

SoggySaussages: Oh sorry, maybe you can’t see my response. I replied to this saying “I’m not sure I understand what you’re asking for,” which still stands.

SoggySaussages:
Sorry I’m just realizing a bunch of comments were left pending in s review instead of posted as individual comments, my apologies. I’m not sure how I did that.

content/docs/custom-commands/commands.md
l-zeuch: I'd like to have this section (after the callout and until the next heading) moved further up, before the table explaining how a cron expression is structured.

As it stands, we have a screenshot of the page, with its appropriate caption, then go to great lengths of explaining a single marked thing in that screenshot (the cron expression labelled 1), *then* have more references to the image. This makes this section awkward to read--a reader would have to scroll back up for the visual reference the image is supposed to provide.

Though these additional fields are quickly explained, I'd rather have them at the top with the description of the cron expression, and only after the image is fully describe we explain how to construct a cron expression. That way, we have the image closer to the text that is describing it, eliminating or reducing the need to scroll back up.

SoggySaussages: That's done.

l-zeuch (approved):

PR #61, opened 2024-12-16
all: lint markdown files

l-zeuch:
Add and configure a markdown linter. While we're at it, also fix the
lints it threw, excluding select files (especially all-commands).

----
This is a combination of two commits. Review should not be required,
though it may be useful to review at least the first commit.

- **workflows: add and configure a Markdown linter**
- **all: lint markdown files**

**Terms**


l-zeuch:
Everything works as expected (though I did test it locally with `act`,
one can never be too sure). Merging as-is.

PR #62, opened 2025-01-01
frontend: add selfhosting view toggle

### LINKED ISSUE #25: [Discussion]: Integrating Info For Self-Hosters Into Documentation
This thread serves to discuss the way going forward on how to integrate infos about selfhosting into the documentation. A transcript of relevant parts of the discussion as started in https://github.com/botlabs-gg/yagpdb-docs-v2/pull/15#discussion_r1597730721 follows.

----
@jo3-l:
> I think it should be made more clear that this section is really only relevant to those self-hosting YAGPDB.

@l-zeuch:
> Personally I think it is better to move anything relevant to selfhosters into its own category, which will mostly be just configuration values by then. This way we can keep the main categories clean and focused on end-user features. WDYT?

@SoggySaussages:
> I'm conflicted. I believe that the documentation should be for all users, including self hosters, hence why I've elected to include extra detail about inner workings of the bot in many places where the end-user wouldn't particularly care. On the other hand, I could see there being a section for end users and a section for hosters. I do think however that I personally prefer having the documentation of a feature thoroughly describe the feature's use and workings with detail and specificity.
>
> When looking at similar bots such as Red, their documentation is primarily aimed at the hoster with detailed descriptions of how the end user would use the features later on. Contrastingly, MEE6 documentation is composed of a series of tutorials walking the end user through setting up each plugin step by step. YAG docs fall somewhere in the middle, but the current version leans more towards the MEE6 docs than the Red docs, which I personally don't love.
>
> I think in an ideal world we have a full documentation and then quick start guides for end-users setting up plugins. Maybe expanding the learn page to all plugins of the bot rather than exclusively custom commands?
>
> I think I could maybe see a subpage for self hosting on each relevant category, but I think putting self hosting anything in its own category, a), with subpages for each plugin would be messy and a waste of space, just mirroring the end user layout but under a self hosting category, or b), putting everything on a single page, requiring users to jump up and down a single page to get to information about several completely unrelated plugins just doesn' sit right with me. With either option we are forcing users to jump back and forth between the end user and self hoster's documentation for the same plugin.

----

Soggy summarised it quite well, there are very valid reasons for both approaches.

A reason for a separate chapter is that we (more or less) clearly signal "this is a separate thing", though the
subliminal message behind that sectioning may be lost on the average user. It might also be lost on me, sometimes.

Another reasoning for this approach is that the entire configuration for a self-hosted YAGPDB instance can fit on one page. That way, everything you need to know *for self-hosting* is in one place, instead of being scattered through the entire document.

That said, however, and thinking intensely about this, I am more inclined to agree with Soggy. If we clearly mark
information for self-hosters as such (as suggested by Joe), maybe even put them at the very end (as is already the case in #15), we should be in the clear.

That way, it is still decently clear that this is a separate thing, but we don't have to maintain *yet another* page. It should also be easier to document new features, because we only have to focus on one page, instead of remembering to add it to a separate self-hosting section.

Please let me know what you think, especially if you're reading this as an end-user. (hi!)

l-zeuch:
Implement a "selfhosting view" that optionally shows information only
relevant to or available on a selfhosted instance of YAGPDB.

Closes #25 -- see there for additional discussion.

**Terms**

### DIFF
--- a/assets/js/custom.js
+++ b/assets/js/custom.js
@@ -1,1 +1,22 @@
 // Put your custom JS code here
+
+// self-hosting view.
+const getSelfhost = () => {
+  return localStorage.getItem('selfhost') || 'no';
+};
+
+const setSelfhost = (selfhost) => {
+  localStorage.setItem('selfhost', selfhost);
+  document.documentElement.setAttribute('data-selfhost', selfhost);
+};
+
+setSelfhost(getSelfhost());
+
+window.addEventListener('DOMContentLoaded', () => {
+  document.querySelectorAll('[data-selfhost-value]').forEach((element) => {
+    element.addEventListener('click', (event) => {
+      const selfhost = element.getAttribute('data-selfhost-value');
+      setSelfhost(selfhost);
+    });
+  }
+)});
--- a/assets/scss/common/_custom.scss
+++ b/assets/scss/common/_custom.scss
@@ -167,6 +167,10 @@ html[data-selfhost="no"] {
   .icon-tabler-server-off {
     display: block;
   }
+
+  .selfhosting {
+    display: none;
+  }
 }
 
 @media (hover: hover) {
--- a/assets/scss/common/_custom.scss
+++ b/assets/scss/common/_custom.scss
@@ -147,6 +147,28 @@ pre.shiki {
   opacity: var(--ec-frm-inlBtnBgActOpa);
 }
 
+// show active server if self-hosting view enabled
+html[data-selfhost="yes"] {
+  .icon-tabler-server {
+    display: block;
+  }
+
+  .icon-tabler-server-off {
+    display: none;
+  }
+}
+
+// hide active server if self-hosting view disabled
+html[data-selfhost="no"] {
+  .icon-tabler-server {
+    display: none;
+  }
+
+  .icon-tabler-server-off {
+    display: block;
+  }
+}
+
 @media (hover: hover) {
   .highlight .copy button {
       opacity: 0;
--- a/config/postcss.config.js
+++ b/config/postcss.config.js
@@ -26,6 +26,7 @@ module.exports = {
 				'data-global-alert',
 				'data-pane', // tabs.js
 				'data-popper-placement',
+				'data-selfhost',
 				'data-sizes',
 				'data-toggle-tab', // tabs.js
 				'id',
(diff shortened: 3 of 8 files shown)


l-zeuch:
Unfortunately the link render hook does not enjoy having to search for
headings within shortcodes; I'm inclined to somehow quench or ignore
those warnings, unless someone more familiar with the link renderer has
an approach to make it **not** complain about that situation.

assets/js/custom.js
```diff
@@ -1 +1,29 @@
...
+const getSelfhost = () => {
+  if (storedSelfhost) {
+    return storedSelfhost;
+  }
+
+  return 'no';
+};
+
+const setSelfhost = function (selfhost) {
+  document.documentElement.setAttribute('data-bs-selfhost', selfhost);
```
jo3-l, line 16: Could we call this data attribute `data-selfhost` (bs = Bootstrap, which this seems orthogonal to?)

assets/js/custom.js
```diff
@@ -1 +1,29 @@
...
+
+const getSelfhost = () => {
+  if (storedSelfhost) {
+    return storedSelfhost;
+  }
+
+  return 'no';
+};
+
+const setSelfhost = function (selfhost) {
```
jo3-l, line 15: This can be `(selfhost) => { ... }` instead of a function literal.

Unrelated, could we also update `localStorage` in this function?

assets/js/custom.js
```diff
@@ -1 +1,29 @@
...
+
+const storedSelfhost = localStorage.getItem('selfhost');
+
+const getSelfhost = () => {
+  if (storedSelfhost) {
+    return storedSelfhost;
+  }
+
+  return 'no';
+};
```
jo3-l, line 13: Even though this function is only called once, I think it's clearest if we always try to fetch from localStorage inside the function: `return localStorage.getItem('selfhost') || 'no';`. Otherwise `getSelfhost` may not accurately reflect the correct state, depending on whether selfhost has been updated since initialization.

assets/js/custom.js
l-zeuch: Thanks for the tip! sure, that sounds reasonable.

assets/js/custom.js
l-zeuch: Absolutely, this is a remnant from the initial phase where I took strong inspiration from the already existing code in doks.

l-zeuch:
I have force-pushed an update; should be good now. Thanks for the review :3

jo3-l (approved):

PR #63, opened 2025-01-11
images: wrap images in a lightbox

l-zeuch:
Wrap images in a lightbox, increasing them to full-screen size when
clicked. There are JavaScript solutions available (e.g. lightbox2)[^0],
though I opted for a purely CSS-based approach.[^1]

[^0]: https://lokeshdhakar.com/projects/lightbox2/
[^1]: https://codesalad.dev/blog/how-to-create-an-image-lightbox-in-pure-css-25

We can further modify the link render hook such that wrapping the image
in question is optional—perhaps this is sensible for rather small
images? What do you think?

Local testing reveals no issues; it may be worth testing on a mobile
device or viewport as well.

**Terms**

### DIFF
--- a/assets/scss/components/_images.scss
+++ b/assets/scss/components/_images.scss
@@ -25,3 +25,40 @@ figcaption {
   margin-top: 0.5rem;
   font-style: italic;
 }
+
+/* CSS-only lightbox. See https://codesalad.dev/blog/how-to-create-an-image-lightbox-in-pure-css-25 */
+.lightbox {
+  /* Default to hidden */
+  display: none;
+
+  /* Overlay entire screen */
+  position: fixed;
+  z-index: 99999;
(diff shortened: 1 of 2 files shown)


l-zeuch:
Sidenote: Perhaps we should also explore [lazy loading](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading), because the implementation requires rendering the image in question twice.
I will implement this in a future commit, potentially alongside the option to not wrap an image in a lightbox.

SoggySaussages:
`-evalcc {{ addReactions "soundsgoodtome:1326889055668932659" }}`

l-zeuch:
I'm going to deploy this now--if necessary, we can disable the lightbox on select images in a future commit.

PR #68, opened 2025-02-20
workflows: use new and maintained mdl implementation

l-zeuch:
The currently used action to lint markdown files does not properly
produce annotations, which would be really useful to have. Use a newer
version that does support annotating files.

**Terms**

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,16 +2,19 @@ name: Lint
 
 on:
   pull_request:
-    paths:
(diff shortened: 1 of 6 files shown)


l-zeuch:
To the lurkers in <#829758329390891029> (`#docs-github-feed`): this implementation doesn't care about modified files either, github just says "annotations in unmodified files" which are real easy to ignore. Nobody has to inspect the error logs anymore and think "what files did I modify again?".

Very. Convenient.

PR #75, opened 2025-05-09
docs/reference: reword time syntax and data explanation

galen8183:
Signed-off-by: Galen CC <galen8183@gmail.com>

**Terms**

### DIFF
--- a/content/docs/reference/templates/syntax-and-data.md
+++ b/content/docs/reference/templates/syntax-and-data.md
@@ -891,9 +891,12 @@ See [`createTicket`](functions#createticket) for creating a ticket.
 
 ## Time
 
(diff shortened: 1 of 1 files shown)


l-zeuch (approved):
Thanks for this. If it's not too much trouble, could you also replace the italics markers (currently underscores) with single asterisks?

galen8183:
> Thanks for this. If it's not too much trouble, could you also replace the italics markers (currently underscores) with single asterisks?

Done -- is this style preferred everywhere or only outside of tables?

l-zeuch:
> Done -- is this style preferred everywhere or only outside of tables?

I'm working towards preferring it everywhere, but don't go out of your way to touch up the entire file now--I have already done some work relating to that in my work tree.

PR #76, opened 2025-06-04
add reprewards information

savage4618:
This pull request adds information about rep rewards that was previously missing from the project.
**Terms**

Signed off by savage4618


content/docs/fun/reputation.md
```diff
@@ -57,6 +57,21 @@ give/remove points._
...
+
+![Overview of Reputation Role Rewards](overview_reuptationRewards.png)
+
+### Create new reputation role
+
+Select a rep point threshold and role to assign when a user receives that many points.
+
+### Existing reputation roles
+
+This will show all existing reputation roles on your guild
```
l-zeuch, line 73:
```suggestion
A listing of all your existing reputation roles with their respective threshold.
```
I'd like this to be a full sentence.

content/docs/fun/reputation.md
```diff
@@ -57,6 +57,21 @@ give/remove points._
 Users with this role cannot participate in the point system which they can be given/taken away points from. This also
 overrides the _Required role to receive/being taken away points from_.
 
+## Reputation roles
```
l-zeuch, line 60: Title Case maybe? Though the header on the control panel isn't, so I'm not sure.

savage4618: yea, I referenced the page specifically for this one and figured if i made it different I'd get a comment about not matching, but I agree, it felt wrong here.

l-zeuch: let's keep it as-is

l-zeuch:
Thanks!

PR #89, opened 2025-10-07
add return info to exec function docs

Qubus0:
add return info to exec function docs

**Terms**


Wolveric:
> If the response isn't captured in a variable, outputs it to chat where the command was triggered.

That isn't a result of the behaviour of those functions though... that goes for any unbound values...

PR #91, opened 2025-11-17
Fix parameter name in cbutton example

mkwarman:
Updated the parameter name from 'custom-id' to 'custom_id' in the cbutton function. I was confused about why I could not trigger my interaction custom command until I realized the custom id key uses an underscore instead of a dash.

**Terms**

- [X] I have read and understood this project's [Contributing Guidelines](CONTRIBUTING.md)


l-zeuch:
Thanks! Interesting how I missed that :thinking:

PR #119, opened 2026-06-11
added documentation for slash custom commands

ashishjh-bst:
**Terms**

### DIFF
--- a/content/docs/custom-commands/commands.md
+++ b/content/docs/custom-commands/commands.md
@@ -405,6 +405,97 @@ This trigger type executes when a role is added to or removed from a member.
 
 - **Execution**: Cannot be triggered via `execCC` or `scheduleUniqueCC`.
 
+##### Slash Command
+
+This trigger type registers a native Discord [slash command](https://support.discord.com/hc/en-us/articles/26501837786775-Slash-Commands-FAQ) with your server.
+The command executes when a member runs `/your-command` in Discord, and any arguments the member supplies are made available to the response.
+
+Because slash commands are an [interaction](/docs/reference/custom-interactions), the command's response is sent as the reply to the interaction.
+You may make the reply ephemeral using the [Defer mode](#defer-mode) below or the [`ephemeralResponse`](/docs/reference/templates/functions#interactions) function, and you can send followups, modals, components, and so on just like other interaction triggers.
+
+###### Command Trigger
+
+For this trigger type, the **Trigger** (**4**) field is the slash command name (the `/name` members type in Discord).
+
+- Maximum 32 characters.
+- Lowercase letters, numbers, dashes (`-`) and underscores (`_`) only.
+- It cannot match the name of a built-in YAGPDB slash command.
+
+###### Description
+
+A slash command requires a **Description** (1–100 characters), shown to members in Discord's command picker.
+
+###### Options
+
+Slash commands may define up to **25 options** (arguments). Each option has a:
+
+- **Name**: 1–32 lowercase characters (letters, numbers, dashes, underscores). Must be unique within the command.
+- **Type**: see the table below.
+- **Description**: 1–100 characters.
+- **Required** toggle: required options are always asked for before optional ones.
+
+| Type          | Description                                                                                  |
+| ------------- | -------------------------------------------------------------------------------------------- |
+| Text          | A free-text string.                                                                          |
+| Text menu     | A string with a fixed set of **choices** the member picks from.                              |
+| Integer       | A whole number.                                                                              |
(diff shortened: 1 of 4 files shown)


content/docs/custom-commands/commands.md
```diff
@@ -405,6 +405,97 @@ This trigger type executes when a role is added to or removed from a member.
...
+| Text menu     | A string with a fixed set of **choices** the member picks from.                              |
+| Integer       | A whole number.                                                                              |
+| Integer menu  | An integer with a fixed set of **choices**.                                                  |
+| Number        | A decimal number.                                                                            |
+| Number menu   | A number with a fixed set of **choices**.                                                    |
+| True/False    | A boolean.                                                                                    |
+| User          | A member of the server.                                                                       |
+| Channel       | A channel in the server.                                                                      |
+| Role          | A role in the server.                                                                         |
+| User or Role  | A mentionable, i.e. either a user or a role.                                                  |
```
l-zeuch: Style nit: the end markers of these rows are not aligned with the rest

content/docs/custom-commands/commands.md
```diff
@@ -405,6 +405,97 @@ This trigger type executes when a role is added to or removed from a member.
...
+| `.Options`        | An `SDict` keyed by option name, holding the value the member supplied for each option. Optional options the member omits are absent. |
+| `.Args`           | An ordered slice of the supplied values, with the **command name at index 0** followed by the provided option values.               |
+| `.CmdArgs`        | The same ordered values as `.Args` but **without** the command name (i.e. `.Args` from index 1 onward).                             |
+| `.CommandName` / `.Cmd` | The slash command name that was run.                                                                                         |
+| `.IsSlashCommand` | `true` when the command was triggered by a slash command.                                                                          |
+| `.Interaction`    | The triggering [interaction](/docs/reference/custom-interactions#parsing-an-interaction) object.                                   |
+| `.InteractionData`| The raw application command interaction data from Discord.                                                                          |
```
l-zeuch: same style nit here

content/docs/custom-commands/commands.md
```diff
@@ -405,6 +405,97 @@ This trigger type executes when a role is added to or removed from a member.
...
+
+Snowflake-typed options are resolved to full objects: a `User` (or chosen-user `User or Role`) option yields a user object, a `Channel` option a channel object, and a `Role` (or chosen-role `User or Role`) option a role object.
```
l-zeuch: Q: Is the "User or Role" option really named as such, or is it "mentionable"?
If the latter, what do you think about introducing the concept of "mentionable" and explaining that a mentionable is either a user or a role?
I *think* we already have something like that explained in interactions someplace, most likely learn.

content/docs/custom-commands/commands.md
```diff
@@ -405,6 +405,97 @@ This trigger type executes when a role is added to or removed from a member.
...
+```yag
+{{ $name := .Options.name }}
+{{ $target := .Options.target }}
+Hello {{ $target.Mention }}, {{ .User.Username }} says: {{ $name }}
+```
+
+{{< callout context="note" title="Note: No source message" icon="outline/info-circle" >}}
+
+A slash command interaction has no source message, so YAGPDB synthesizes a minimal one.
+`.User`, `.Member`, and member-dependent functions such as `addRoleID` work as usual, but message-specific data such as `.Message.Content` is not meaningful.
```
l-zeuch: Q: Is not meaningful... and therefore not populated, or is it populated but has nothing useful (e.g. empty content) available?

This could be clarified, I think.

l-zeuch (commented):
Thanks for the work on both the feature and the docs! I have a few comments, see below.

content/docs/custom-commands/commands.md
ashishjh-bst: I made it a User or a Role to make it easy for people to understand what it does, though I don't mind changing it to Mentionable in the UI.

content/docs/custom-commands/commands.md
ashishjh-bst: I'll change it to `".Message.Content" doesn't exist, as there is no underlying message associated with the trigger. `

content/docs/custom-commands/commands.md
l-zeuch: I was just wondering, but I agree that "mentionable" can be confusing. Let's keep it as-is.

content/docs/custom-commands/commands.md
l-zeuch: Sounds good to me!

l-zeuch (approved):
Many thanks!

